PR #2, opened 2023-09-07
Bootcamp submission - Ethan Ahn

Ethan118:


modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -69,12 +69,25 @@ def run(self,
...
         # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
         # ============
 
         return command
+
+    def getRelativeDistance(self, report: drone_report.DroneReport) -> float:
+        """
+        Returns the distance between the drone and the waypoint.
+        """
+        return ((self.waypoint.location_x - report.position.location_x)**2 + (self.waypoint.location_y - report.position.location_y)**2)**0.5
```
Xierumeng, line 93: Square roots are computationally expensive, would it be possible to avoid using it? (You do not need to change your bootcamp, just something to think about).

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,13 +68,57 @@ def run(self,
...
         return command
+    
+    def findNearestLandingPad(self, position: location.Location, landing_pad_locations: "list[location.Location]") -> location.Location:
+        """
+        Returns the nearest landing pad to the drone.
+        """
+        nearestLandingPad = landing_pad_locations[0]
+        nearestDistance = self.getRelativeDistance(position, nearestLandingPad)
+        for landingPad in landing_pad_locations:
+            distance = self.getRelativeDistance(position, landingPad)
```
Xierumeng, line 114: Calculating distance twice unnecessarily. Is there a better way? (You do not need to change your bootcamp, just something to think about).

modules/bootcamp/detect_landing_pad.py
```diff
@@ -84,29 +84,33 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+        predictions = self.__model.predict(source=image, conf=0.7, device=self.__DEVICE, verbose=False)
 
         # Get the Result object
-        prediction = ...
+        prediction = predictions[0]
 
         # Plot the annotated image from the Result object
         # Include the confidence value
-        image_annotated = ...
+        image_annotated = prediction.plot()
```
Xierumeng, line 94: Does this image with bounding boxes show the confidence of the detection?

modules/bootcamp/detect_landing_pad.py
```diff
@@ -84,29 +84,33 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+        boxes_cpu = boxes_xyxy.detach().cpu().numpy()
 
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
         # Hint: .shape gets the dimensions of the numpy array
-        # for i in range(0, ...):
+        for i in range(0, boxes_cpu.shape[0]):
             # Create BoundingBox object and append to list
-            # result, box = ...
+            box = bounding_box.BoundingBox.create(boxes_cpu[i])[1]
```
Xierumeng, line 109: What if box is `None` ?

modules/bootcamp/detect_landing_pad.py
```diff
@@ -84,29 +84,33 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+            bounding_boxes.append(box)
+
+        # Return the list of bounding boxes and the annotated image
+        return bounding_boxes, image_annotated
```
Xierumeng, line 113: You should remove `raise NotImplementedError` since you have a working implementation.

Xierumeng:
Looks pretty good! A couple of comments to address.

modules/bootcamp/detect_landing_pad.py
```diff
@@ -84,33 +84,35 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
         # Hint: .shape gets the dimensions of the numpy array
-        # for i in range(0, ...):
+        for i in range(0, boxes_cpu.shape[0]):
             # Create BoundingBox object and append to list
-            # result, box = ...
+            box = bounding_box.BoundingBox.create(boxes_cpu[i])[1]
+            if (box is not None):
+                bounding_boxes.append(box)
```
Xierumeng, line 111: The pattern used often in Autonomy is a return of `(bool, DataStructure | None)` , where `bool` is the result and `DataStructure | None` is possible value (similar to C++'s `std::optional` ). It would be better to check the result than whether the value exists (there are some cases where returning `True, None` and `False, 0` might be allowed).

You can use tuple assignment to avoid having to manually unpack the tupled returned from `.create()` .

Xierumeng, line 111: If 1 bounding box is bad, we might not want the other bounding boxes (the reason for this is from statistics/data collection).

modules/bootcamp/detect_landing_pad.py
```diff
@@ -84,33 +84,39 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Hint: .shape gets the dimensions of the numpy array
-        # for i in range(0, ...):
+        for i in range(0, boxes_cpu.shape[0]):
             # Create BoundingBox object and append to list
-            # result, box = ...
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+            if result:
+                bounding_boxes.append(box)
+            else:
+                print("Failed to create bounding box")
```
Xierumeng, line 113: Logging is good! It's unfortunate that the subteam still doesn't have a good logger (1 of the open tasks we have right now...). (You do not need to change anything here).

Xierumeng (commented):
LGTM.

PR #3, opened 2023-09-09
Mihir Bootcamp

mgupta27:
Some suggestions we could do to improve bootcamp experience.

Task 4
* What do we do if there are 0 landing pads to be found (very unlikely but still a possibility)
* Can the string representation of Location be changed to (x, y)
    * Reading locations like this can be hard when debugging: [36.54166774947024,49.630434876450444, 46.36469682254967,45.95929128107284, 50.976307743773624,50.275420838364504]

### DIFF
--- a/modules/bootcamp/decision_simple_waypoint.py
+++ b/modules/bootcamp/decision_simple_waypoint.py
@@ -69,9 +69,20 @@ class DecisionSimpleWaypoint(base_decision.BaseDecision):
         # ============
 
         # Do something based on the report and the state of this class...
+        if report.status == drone_status.DroneStatus.HALTED and report.position != self.waypoint:
+            waypoint_relative_to_drone_x = self.waypoint.location_x - report.position.location_x
+            waypoint_relative_to_drone_y = self.waypoint.location_y - report.position.location_y
+
+            command = commands.Command.create_set_relative_destination_command(
+                waypoint_relative_to_drone_x,
+                waypoint_relative_to_drone_y,
+            )
+
+        if report.status == drone_status.DroneStatus.HALTED and report.position == self.waypoint:
+            command = commands.Command.create_land_command()
 
         # Remove this when done
-        raise NotImplementedError
+        # raise NotImplementedError
 
         # ============
         # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
--- a/modules/bootcamp/decision_simple_waypoint.py
+++ b/modules/bootcamp/decision_simple_waypoint.py
@@ -87,9 +87,6 @@ class DecisionSimpleWaypoint(base_decision.BaseDecision):
             # Create land command
             command = commands.Command.create_land_command()
 
-        # Remove this when done
-        # raise NotImplementedError
-
         # ============
         # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
         # ============
--- a/modules/bootcamp/decision_simple_waypoint.py
+++ b/modules/bootcamp/decision_simple_waypoint.py
@@ -69,16 +69,22 @@ class DecisionSimpleWaypoint(base_decision.BaseDecision):
         # ============
 
         # Do something based on the report and the state of this class...
+
+        # When drone is HALTED and the drone is not at the waypoint then move towards waypoint
         if report.status == drone_status.DroneStatus.HALTED and report.position != self.waypoint:
+            # Get relative x and y position of waypoint to drone
             waypoint_relative_to_drone_x = self.waypoint.location_x - report.position.location_x
             waypoint_relative_to_drone_y = self.waypoint.location_y - report.position.location_y
 
+            # Create relative destination command
             command = commands.Command.create_set_relative_destination_command(
                 waypoint_relative_to_drone_x,
                 waypoint_relative_to_drone_y,
             )
 
+        # When drone is HALTED and the drone is at the desired waypoint then land
         if report.status == drone_status.DroneStatus.HALTED and report.position == self.waypoint:
+            # Create land command
             command = commands.Command.create_land_command()
 
         # Remove this when done
--- a/modules/bootcamp/decision_simple_waypoint.py
+++ b/modules/bootcamp/decision_simple_waypoint.py
@@ -70,8 +70,11 @@ class DecisionSimpleWaypoint(base_decision.BaseDecision):
 
         # Do something based on the report and the state of this class...
 
-        # When drone is HALTED and the drone is not at the waypoint then move towards waypoint
-        if report.status == drone_status.DroneStatus.HALTED and report.position != self.waypoint:
+        distance_to_waypoint = self.distance_between_points(self.waypoint, report.position)
+
+        if report.status == drone_status.DroneStatus.HALTED and distance_to_waypoint > self.acceptance_radius:
+            # When drone is HALTED and the drone is not near the waypoint then move towards waypoint
+
             # Get relative x and y position of waypoint to drone
             waypoint_relative_to_drone_x = self.waypoint.location_x - report.position.location_x
             waypoint_relative_to_drone_y = self.waypoint.location_y - report.position.location_y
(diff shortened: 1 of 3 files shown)


modules/bootcamp/detect_landing_pad.py
```diff
@@ -84,32 +85,41 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         bounding_boxes = []
         # Hint: .shape gets the dimensions of the numpy array
-        # for i in range(0, ...):
+        for i in range(0, boxes_cpu.shape[0]):
             # Create BoundingBox object and append to list
-            # result, box = ...
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+
+            if not result:
+                continue
```
Xierumeng, line 117: If 1 bounding box is bad, we might not want the other bounding boxes (the reason for this is from statistics/data collection).

modules/bootcamp/detect_landing_pad.py
Xierumeng, line 1: Looking good!

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -70,8 +70,22 @@ def run(self,
 
         # Do something based on the report and the state of this class...
 
-        # Remove this when done
-        raise NotImplementedError
+        # When drone is HALTED and the drone is not at the waypoint then move towards waypoint
+        if report.status == drone_status.DroneStatus.HALTED and report.position != self.waypoint:
```
Xierumeng, line 74: What if the drone and waypoint are off by a tiny amout (e.g. 0.0001)? Better to avoid floating point equality comparisons and use a tolerance instead.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -5,7 +5,7 @@
 """
 # Disable for bootcamp use
 # pylint: disable=unused-import
-
+import math
```
Xierumeng, line 8: Bootcamp is completable without imports.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,12 +72,61 @@ def run(self,
...
+            command = commands.Command.create_set_relative_destination_command(
+                waypoint_relative_to_drone_x,
+                waypoint_relative_to_drone_y,
+            )
+
+        # When drone is HALTED and the drone is at the desired waypoint then find closest landing pad
+        # and move towards selected landing pad location 
+        if report.status == drone_status.DroneStatus.HALTED and report.position == self.waypoint:
+            landing_pad_distances = []
+            for i in range(0, len(landing_pad_locations)):
```
Xierumeng, line 90: What about using an iterator: `for landing_pad_location in landing_pad_locations:`

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,12 +72,61 @@ def run(self,
...
+            command = commands.Command.create_set_relative_destination_command(
+                closest_pad_relative_to_drone_x,
+                closest_pad_relative_to_drone_y,
+            )
+
+            self.landing_pad = closest_landing_pad_location
+
+        # When drone is HALTED and the drone is at the desired landing pad then land
+        if self.landing_pad is not None and report.position == self.landing_pad:
+            command = commands.Command.create_land_command()
```
Xierumeng, line 116: What happens if the drone is still moving (e.g. it overshoots slightly)? Would this be a valid command?

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,12 +72,61 @@ def run(self,
         # ============
 
         # Do something based on the report and the state of this class...
+        
+        # When drone is HALTED and the drone is not at the waypoint then move towards waypoint
```
Xierumeng, line 76: What if the drone was moving towards the landing pad and then something caused it to suddenly halt (i.e. it decided for some reason to be in the halted state, interrupting the movement)?

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -69,12 +69,37 @@ def run(self,
...
+            # Create halt command
+            command = commands.Command.create_halt_command()
 
         # ============
         # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
         # ============
 
         return command
+
+    @staticmethod
```
Xierumeng, line 100: Good use of static method!

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -69,12 +69,37 @@ def run(self,
...
+    def distance_between_points(point_1: location.Location, point_2: location.Location) -> float:
```
Xierumeng, line 101: Good use of type hinting!

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -69,12 +69,37 @@ def run(self,
...
+        distance_x = point_1.location_x - point_2.location_x
+        distance_y = point_1.location_y - point_2.location_y
+
+        return max(abs(distance_x), abs(distance_y))
```
Xierumeng, line 105: Explain in a comment why this is the formula being used (as opposed to the regular `sqrt(sum of (component^2))` ).

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,12 +71,57 @@ def run(self,
         # ============
 
         # Do something based on the report and the state of this class...
-
-        # Remove this when done
-        raise NotImplementedError
+        if report.status == drone_status.DroneStatus.HALTED:
+            if self.state == 0:
+                # If the drone is near the destination then set destination to closest landing pad
```
Xierumeng, line 76: Current destination?

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,12 +71,57 @@ def run(self,
...
+                if self.distance_between_points(self.destination, report.position) <= self.acceptance_radius:
+                    # Calculate the distances of the detected landing pad
+                    landing_pad_distances = []
+                    for landing_pad_location in landing_pad_locations:
+                        distance = self.distance_between_points(landing_pad_location, report.position)
```
Xierumeng, line 81: Is this the correct distance formula? What if these are the landing pads:
* (1,0) to the right
* (0.9,0.9) to the upper right

Would the drone fly to the correct landing pad?

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,12 +71,57 @@ def run(self,
...
+                        landing_pad_distances.append(distance)
+
+                    # Find the landing pad closest to the drone
+                    shortest_distance = landing_pad_distances[0]
+                    closest_landing_pad_index = 0
```
Xierumeng, line 86: While the bootcamp specifies at least 1 landing pad, what happens if there were no landing pads? (You do not need to change anything here, just something to think about).

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,12 +71,57 @@ def run(self,
...
+
+                    for i in range(1, len(landing_pad_distances)):
```
Xierumeng, line 88: Prefer iteration (e.g. `for distance in landing_pad_distances[...]` ). You can use slicing to exclude the 1st item.

Xierumeng (commented):
Reviewed.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,12 +71,54 @@ def run(self,
...
+                        distance = self.distance_between_points(landing_pad_location, report.position)
+                        landing_pad_distances.append(distance)
+
+                    # Find the landing pad closest to the drone
+                    shortest_distance = landing_pad_distances[0]
+
+                    for i, landing_pad_distance in enumerate(landing_pad_distances[1:]):
+                        if landing_pad_distance < shortest_distance:
+                            shortest_distance = landing_pad_distance
+                            self.destination = landing_pad_locations[i]
```
Xierumeng, line 90: What if the closest landing pad is the first in the list? Would this be set correctly?

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,12 +71,54 @@ def run(self,
...
         # ============
 
         return command
+
+    @staticmethod
+    def distance_between_points(point_1: location.Location, point_2: location.Location) -> float:
+        distance_x = point_1.location_x - point_2.location_x
+        distance_y = point_1.location_y - point_2.location_y
+
+        return (distance_x ** 2 + distance_y ** 2) ** 0.5
```
Xierumeng, line 124: Square roots are computationally expensive. Is there a way to avoid using it? (You do not need to change anything here, just something to think about).

Xierumeng (commented):
LGTM.

PR #11, opened 2023-09-14
Himank Bootcamp PR

steadyfall:
I have completed all the 4 given tasks and have created a PR, as per the bootcamp docs

### DIFF
--- a/modules/bootcamp/decision_simple_waypoint.py
+++ b/modules/bootcamp/decision_simple_waypoint.py
@@ -37,11 +37,42 @@ class DecisionSimpleWaypoint(base_decision.BaseDecision):
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
-        # Add your own
+        self.action_dict = dict(zip(("MOVE", "HALT", "LAND"), range(3,6)))
 
         # ============
         # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
         # ============
+
+    def distance_between_waypoint(self, given_location:location.Location) -> bool:
+        """
+        Returns the distance between the given location and waypoint.
+        """
+        given_loc_from_waypoint_location_x = self.waypoint.location_x - given_location.location_x 
+        given_loc_from_waypoint_location_y = self.waypoint.location_y - given_location.location_y
+        return (given_loc_from_waypoint_location_x, given_loc_from_waypoint_location_y)
+    
+    def check_if_near_waypoint(self, given_location:location.Location) -> bool:
+        """
+        Checks if the given location is on the waypoint by an acceptance radius.
+        """
+        absolute_acceptance_radius = abs(self.acceptance_radius)
+        difference_location_x, difference_location_y = self.distance_between_waypoint(given_location)
+        if abs(difference_location_x) < absolute_acceptance_radius and abs(difference_location_y) < absolute_acceptance_radius:
+            return True
+        return False
+    
+    def next_relative_coordinates_to_waypoint(self, given_location:location.Location) -> "tuple[float]":
+        """
+        Returns the relative x and y coordinates for drone to be sent to.
+        """
+        relative_x, relative_y = self.distance_between_waypoint(given_location)
+        divider = 1
+        if abs(relative_x) > abs(relative_y):
+            return (relative_x/divider, relative_y)
+        elif abs(relative_x) < abs(relative_y):
+            return (relative_x, relative_y/divider)
+        else:
+            return (relative_x/divider, relative_y/divider)
 
     def run(self,
             report: drone_report.DroneReport,
@@ -67,11 +98,33 @@ class DecisionSimpleWaypoint(base_decision.BaseDecision):
         # ============
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
+     
+        action = None
+        report_status = report.status
+        report_position = report.position
 
-        # Do something based on the report and the state of this class...
+        if report_status == drone_status.DroneStatus.LANDED:
+            action = None
+        elif report_status == drone_status.DroneStatus.HALTED:
+            if self.check_if_near_waypoint(report_position):
+                action = self.action_dict["LAND"]
+            else:
+                action = self.action_dict["MOVE"]
+        """ elif report_status == drone_status.DroneStatus.MOVING:
+            action = self.action_dict["HALT"] """
+
+        if action is None:
+            pass
+        elif action == self.action_dict["MOVE"]:
+            relative_x, relative_y = self.next_relative_coordinates_to_waypoint(report_position)
+            command = commands.Command.create_set_relative_destination_command(relative_x, relative_y)
+        elif action == self.action_dict["HALT"]:
+            command = commands.Command.create_halt_command()
+        elif action == self.action_dict["LAND"]:
+            command = commands.Command.create_land_command()
 
         # Remove this when done
-        raise NotImplementedError
+        # raise NotImplementedError
 
         # ============
         # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
--- a/modules/bootcamp/decision_simple_waypoint.py
+++ b/modules/bootcamp/decision_simple_waypoint.py
@@ -47,10 +47,10 @@ class DecisionSimpleWaypoint(base_decision.BaseDecision):
         """
         Returns the distance between the given location and waypoint.
         """
-        given_loc_from_waypoint_location_x = self.waypoint.location_x - given_location.location_x 
+        given_loc_from_waypoint_location_x = self.waypoint.location_x - given_location.location_x
         given_loc_from_waypoint_location_y = self.waypoint.location_y - given_location.location_y
         return (given_loc_from_waypoint_location_x, given_loc_from_waypoint_location_y)
-    
+
     def check_if_near_waypoint(self, given_location:location.Location) -> bool:
         """
         Checks if the given location is on the waypoint by an acceptance radius.
@@ -60,7 +60,7 @@ class DecisionSimpleWaypoint(base_decision.BaseDecision):
         if abs(difference_location_x) < absolute_acceptance_radius and abs(difference_location_y) < absolute_acceptance_radius:
             return True
         return False
-    
+
     def next_relative_coordinates_to_waypoint(self, given_location:location.Location) -> "tuple[float]":
         """
         Returns the relative x and y coordinates for drone to be sent to.
@@ -98,7 +98,7 @@ class DecisionSimpleWaypoint(base_decision.BaseDecision):
         # ============
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
-     
+
         action = None
         report_status = report.status
         report_position = report.position
--- a/modules/bootcamp/decision_waypoint_landing_pads.py
+++ b/modules/bootcamp/decision_waypoint_landing_pads.py
@@ -37,12 +37,68 @@ class DecisionWaypointLandingPads(base_decision.BaseDecision):
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
-        # Add your own
+        self.action_dict = dict(zip(("MOVE", "HALT", "LAND"), range(3,6)))
+        self.origin = location.Location(0,0)
 
         # ============
         # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
         # ============
 
+    def relative_coordinates_of_target(self, target_location:location.Location, given_location:location.Location) -> bool:
+        """
+        Returns the relative coordinates of target w.r.t given location.
+        """
+        relative_location_x = target_location.location_x - given_location.location_x 
+        relative_location_y = target_location.location_y - given_location.location_y
+        return (relative_location_x, relative_location_y)
+    
+    def check_if_near_target(self, target_location:location.Location, given_location:location.Location) -> bool:
+        """
+        Checks if the given location is near the target by an acceptance radius.
+        """
+        absolute_acceptance_radius = abs(self.acceptance_radius)
+        difference_location_x, difference_location_y = self.relative_coordinates_of_target(target_location, given_location)
+        if abs(difference_location_x) < absolute_acceptance_radius and abs(difference_location_y) < absolute_acceptance_radius:
+            return True
+        return False
+    
+    def next_relative_coordinates_to_target(self, 
+                                            target_location:location.Location, 
+                                            given_location:location.Location) -> "tuple[float]":
+        """
+        Returns the relative x and y coordinates for drone to be sent to.
+        """
+        relative_x, relative_y = self.relative_coordinates_of_target(target_location, given_location)
+        divider = 1
+        if abs(relative_x) > abs(relative_y):
+            return (relative_x/divider, relative_y)
+        elif abs(relative_x) < abs(relative_y):
+            return (relative_x, relative_y/divider)
+        else:
+            return (relative_x/divider, relative_y/divider)
+
+    def closest_landing_pad(self, 
+                            given_location:location.Location, 
+                            landing_pad_locations: "list[location.Location]") -> location.Location:
+        """
+        Finds out the closest landing pad from the given location by checking out their distances.
+        """
+        def shortest_distance(target_location:location.Location, given_location:location.Location) -> float:
+            """
+            Finds out the shortest distance between two given locations.
+            """
+            x_1, y_1 = given_location.location_x, given_location.location_y
+            x_2, y_2 = target_location.location_x, target_location.location_y
+            x_square = (x_2 - x_1) ** 2
+            y_square = (y_2 - y_1) ** 2
+            return (x_square + y_square) ** 0.5
+        landing_pad_distances = list(map(lambda loc: shortest_distance(loc, given_location), landing_pad_locations))
+        landing_pad_hashmap = dict(zip(landing_pad_locations, landing_pad_distances))
+        landing_pad_hashmap = tuple(sorted(list(landing_pad_hashmap.items()), key=lambda i: i[1]))
+        return landing_pad_hashmap[0][0]
+    
+
+    
     def run(self,
             report: drone_report.DroneReport,
             landing_pad_locations: "list[location.Location]") -> commands.Command:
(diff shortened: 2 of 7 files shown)


modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,38 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # * verbose
-        predictions = ...
+        predictions = self.__model.predict(source=image,
+                                           conf=0.7,
+                                           device=self.__DEVICE,
+                                           verbose=False)
 
         # Get the Result object
-        prediction = ...
+        prediction = predictions.__getitem__(0)
```
Xierumeng, line 95: Double underscore in Python denotes that it is a private method. Since `predictions` is a list, you can use the square bracket operator instead.

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,38 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
 
         # Plot the annotated image from the Result object
         # Include the confidence value
-        image_annotated = ...
+        image_annotated = prediction.plot(pil=True, conf=True)
```
Xierumeng, line 99: The expected image is a numpy ndarray, not a PIL object.

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,38 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
 
         # Get the xyxy boxes list from the Boxes object in the Result object
-        boxes_xyxy = ...
+        boxes_xyxy = prediction.boxes.xyxy
 
         # Detach the xyxy boxes to make a copy,
         # move the copy into CPU space,
         # and convert to a numpy array
-        boxes_cpu = ...
+        boxes_cpu = prediction.boxes.cpu().xyxy.numpy()
```
Xierumeng, line 107: Use `boxes_xyxy` . Also detach before moving to the CPU.

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,38 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
         # Hint: .shape gets the dimensions of the numpy array
-        # for i in range(0, ...):
+        for i in range(0, boxes_cpu.shape[0]):
             # Create BoundingBox object and append to list
-            # result, box = ...
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+            if result:
+                bounding_boxes.append(box)
```
Xierumeng, line 116: In statistics, it is preferred to remove the whole data point if there is an issue (here, the data point is an image). (You do not need to change anything here, just something to think about).

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,10 +99,32 @@ def run(self,
...
+
+        if report_status == drone_status.DroneStatus.LANDED:
+            action = None
+        elif report_status == drone_status.DroneStatus.HALTED:
+            if self.check_if_near_waypoint(report_position):
+                action = self.action_dict["LAND"]
+            else:
+                action = self.action_dict["MOVE"]
+        """ elif report_status == drone_status.DroneStatus.MOVING:
+            action = self.action_dict["HALT"] """
```
Xierumeng, line 114: Remove.

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -37,12 +37,43 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
...
         # ============
 
-        # Add your own
+        self.action_dict = dict(zip(("MOVE", "HALT", "LAND"), range(3,6)))
 
         # ============
         # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
         # ============
 
+    def distance_between_waypoint(self, given_location:location.Location) -> bool:
```
Xierumeng, line 46: This doesn't return bool?

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -37,12 +37,43 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
...
+        """
+        Checks if the given location is on the waypoint by an acceptance radius.
+        """
+        absolute_acceptance_radius = abs(self.acceptance_radius)
+        difference_location_x, difference_location_y = self.distance_between_waypoint(given_location)
+        if abs(difference_location_x) < absolute_acceptance_radius and abs(difference_location_y) < absolute_acceptance_radius:
+            return True
+        return False
+
+    def next_relative_coordinates_to_waypoint(self, given_location:location.Location) -> "tuple[float]":
```
Xierumeng, line 64: Seems to return `tuple[float, float]` .

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -37,12 +37,43 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
...
+        """
+        Returns the relative x and y coordinates for drone to be sent to.
+        """
+        relative_x, relative_y = self.distance_between_waypoint(given_location)
+        divider = 1
```
Xierumeng, line 69: This doesn't do anything.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -37,12 +37,67 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
-        # Add your own
+        self.action_dict = dict(zip(("MOVE", "HALT", "LAND"), range(3,6)))
+        self.origin = location.Location(0,0)
```
Xierumeng, line 41: What if the drone didn't start at (0,0) , but could start anywhere?

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -37,12 +37,67 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
...
+            Finds out the shortest distance between two given locations.
+            """
+            x_1, y_1 = given_location.location_x, given_location.location_y
+            x_2, y_2 = target_location.location_x, target_location.location_y
+            x_square = (x_2 - x_1) ** 2
+            y_square = (y_2 - y_1) ** 2
+            return (x_square + y_square) ** 0.5
+        landing_pad_distances = list(map(lambda loc: shortest_distance(loc, given_location), landing_pad_locations))
+        landing_pad_hashmap = dict(zip(landing_pad_locations, landing_pad_distances))
+        landing_pad_hashmap = tuple(sorted(list(landing_pad_hashmap.items()), key=lambda i: i[1]))
```
Xierumeng, line 97: This is very convoluted and creates many new objects that will need to be destroyed. Is there a more space efficient way?

Also, sorting is not very efficient. It's possible to get the minimum without sorting the list of landing pad locations.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +123,53 @@ def run(self,
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
-        # Do something based on the report and the state of this class...
+        def shortest_distance(target_location:location.Location, given_location:location.Location) -> float:
```
Xierumeng, line 126: This is duplicate, you can create a static method that can be called instead.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -37,12 +37,67 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
...
+        """
+        def shortest_distance(target_location:location.Location, given_location:location.Location) -> float:
+            """
+            Finds out the shortest distance between two given locations.
+            """
+            x_1, y_1 = given_location.location_x, given_location.location_y
+            x_2, y_2 = target_location.location_x, target_location.location_y
+            x_square = (x_2 - x_1) ** 2
+            y_square = (y_2 - y_1) ** 2
+            return (x_square + y_square) ** 0.5
```
Xierumeng, line 94: Square roots are computationally expensive. Is there another way? (You do not need to change anything here, just something to think about.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +123,53 @@ def run(self,
...
+
+        action = None
+        report_status = report.status
+        report_position = report.position
+        target = self.waypoint
+
+        if report_status in (drone_status.DroneStatus.LANDED, drone_status.DroneStatus.MOVING):
+            pass
```
Xierumeng, line 142: Redundant, can remove.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +123,53 @@ def run(self,
...
+            landing_pad = self.closest_landing_pad(self.waypoint, landing_pad_locations)
+            action = self.action_dict["MOVE"]
+            if self.check_if_near_target(self.waypoint, report_position):
+                target = landing_pad
+            if (self.check_if_near_target(landing_pad, report_position)
+                and (shortest_distance(self.waypoint, report_position) - shortest_distance(self.waypoint, landing_pad) < 0.1)
+                and (
+                    (self.check_if_near_target(landing_pad, self.origin) and self.check_if_near_target(self.waypoint, self.origin))
+                    or (not self.check_if_near_target(report_position, self.origin))
+                )):
```
Xierumeng, line 153: This is also a bit complex. What about checking whether the drone has already arrived at the waypoint (e.g. storing some state)?

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,10 +99,32 @@ def run(self,
...
+            relative_x, relative_y = self.next_relative_coordinates_to_waypoint(report_position)
+            command = commands.Command.create_set_relative_destination_command(relative_x, relative_y)
+        elif action == self.action_dict["HALT"]:
+            command = commands.Command.create_halt_command()
+        elif action == self.action_dict["LAND"]:
+            command = commands.Command.create_land_command()
 
         # Remove this when done
-        raise NotImplementedError
+        # raise NotImplementedError
```
Xierumeng, line 127: Remove.

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,38 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
 
         # Remove this when done
-        raise NotImplementedError
+        # raise NotImplementedError
```
Xierumeng, line 119: Remove.

modules/bootcamp/detect_landing_pad.py
steadyfall: What do you mean by "detach"?

modules/bootcamp/detect_landing_pad.py
steadyfall: Does that mean removing the entire list of bounding boxes if even one bounding box has an issue?

modules/bootcamp/decision_waypoint_landing_pads.py
steadyfall: Then, in that case, wouldn't we need to keep track of the initial position of the drone?

modules/bootcamp/decision_simple_waypoint.py
steadyfall: Removed in commit https://github.com/steadyfall/autonomy-bootcamp-2023/commit/b52182412d693ebd8bd80219d25828199b7f1405.

modules/bootcamp/decision_simple_waypoint.py
steadyfall: Fixed return type to `tuple[float, float]` in commit https://github.com/steadyfall/autonomy-bootcamp-2023/commit/b52182412d693ebd8bd80219d25828199b7f1405.

modules/bootcamp/decision_simple_waypoint.py
steadyfall: Fixed return type to `tuple[float, float]` in https://github.com/steadyfall/autonomy-bootcamp-2023/commit/b52182412d693ebd8bd80219d25828199b7f1405.

modules/bootcamp/decision_simple_waypoint.py
steadyfall: Removed redundant `divider`, simplified function in https://github.com/steadyfall/autonomy-bootcamp-2023/commit/b52182412d693ebd8bd80219d25828199b7f1405.

modules/bootcamp/decision_simple_waypoint.py
steadyfall: Removed in https://github.com/steadyfall/autonomy-bootcamp-2023/commit/68516332c8c11ac19758d028fb9482f3e922ddba.

modules/bootcamp/detect_landing_pad.py
steadyfall: Removed in https://github.com/steadyfall/autonomy-bootcamp-2023/commit/68516332c8c11ac19758d028fb9482f3e922ddba.

modules/bootcamp/decision_waypoint_landing_pads.py
steadyfall: Fixed this with initializing `closest_location` variable with first location in the list and doing a `for` loop for the rest of the list to replace the closer location with `closest_location` in https://github.com/steadyfall/autonomy-bootcamp-2023/commit/f42441a32500e2e48feeab7c6e42da052331244b.

modules/bootcamp/decision_waypoint_landing_pads.py
steadyfall: Created a static method `DecisionWaypointLandingPads.shortest_distance` in https://github.com/steadyfall/autonomy-bootcamp-2023/commit/f42441a32500e2e48feeab7c6e42da052331244b

modules/bootcamp/decision_waypoint_landing_pads.py
steadyfall: I do not square root the number and only square root it when required?

modules/bootcamp/decision_waypoint_landing_pads.py
steadyfall: > This is also a bit complex. What about checking whether the drone has already arrived at the waypoint (e.g. storing some state)?

I had tried doing that, but since this method runs in an infinite loop, any data that I try to store goes away.

modules/bootcamp/decision_waypoint_landing_pads.py
steadyfall: Removed in https://github.com/steadyfall/autonomy-bootcamp-2023/commit/20f63956e807a0a9e6a4d53daa668ca143a44562.

modules/bootcamp/detect_landing_pad.py
steadyfall: According to the [Ultralytics documentation](https://docs.ultralytics.com/modes/predict/#plotting-results), setting `img = True` would return `numpy.ndarray` type, but that was not the case. 

When I tried running pytest with `pil=True, conf=True`, all 3 tests passed, whereas running pytest with `img=True, conf=True` failed all 3 tests.

The error message was the following: `AttributeError: 'bool' object has no attribute 'data'`.

I do not know whether this is a bug or something else.

modules/bootcamp/detect_landing_pad.py
steadyfall: Replaced private method with square bracket operator in https://github.com/steadyfall/autonomy-bootcamp-2023/commit/b6dbf2332d9d15b39d7c58d6abbd692b769e189d.

modules/bootcamp/detect_landing_pad.py
Xierumeng: No, the `img` parameter's type itself is expected to be `np.ndarray` . This is so that you can pass in an existing image that Ultralytics will draw over.

If you leave `pil=True` , what is `image_annotated` object? Do the test output images look fine?

modules/bootcamp/detect_landing_pad.py
Xierumeng: Correct, an empty list would be returned instead. But you don't need to change anything.

modules/bootcamp/detect_landing_pad.py
Xierumeng: Pytorch tensors have a `detach()` method, use that.

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -37,12 +37,37 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
...
+        if abs(difference_location_x) < absolute_acceptance_radius and abs(difference_location_y) < absolute_acceptance_radius:
+            return True
+        return False
+
+    def next_relative_coordinates_to_waypoint(self, given_location:location.Location) -> "tuple[float, float]":
+        """
+        Returns the relative x and y coordinates for drone to be sent to.
+        """
+        relative_x, relative_y = self.distance_between_waypoint(given_location)
+        return (relative_x, relative_y)
```
Xierumeng, line 69: Remove unneeded parantheses.

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -37,12 +37,37 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
-        # Add your own
+        self.action_dict = dict(zip(("MOVE", "HALT", "LAND"), range(3,6)))
```
Xierumeng, line 40: Can use `enum` instead.

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,10 +93,28 @@ def run(self,
...
+        if report_status == drone_status.DroneStatus.LANDED:
+            action = None
+        elif report_status == drone_status.DroneStatus.HALTED:
+            if self.check_if_near_waypoint(report_position):
+                action = self.action_dict["LAND"]
+            else:
+                action = self.action_dict["MOVE"]
+
+        if action is None:
+            pass
```
Xierumeng, line 109: Redundant, remove.

modules/bootcamp/decision_waypoint_landing_pads.py
Xierumeng: Yes. Alternatively, don't use the initial position of the drone.

modules/bootcamp/decision_waypoint_landing_pads.py
Xierumeng: That just procrastinating the work, it doesn't eliminate it. What would happen if there was no square root?

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -37,12 +37,72 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
...
-        # Add your own
+        self.action_dict = dict(zip(("MOVE", "HALT", "LAND"), range(3,6)))
+        self.origin = location.Location(0,0)
 
         # ============
         # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
         # ============
 
+    @staticmethod
+    def shortest_distance(target_location:location.Location, given_location:location.Location) -> float:
```
Xierumeng, line 48: Good use of static method!

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -37,12 +37,72 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
...
+            """
+            Finds out the shortest distance between two given locations.
+            """
+            x_1, y_1 = given_location.location_x, given_location.location_y
+            x_2, y_2 = target_location.location_x, target_location.location_y
+            x_square = (x_2 - x_1) ** 2
+            y_square = (y_2 - y_1) ** 2
+            return (x_square + y_square) ** 0.5
+    
+    def relative_coordinates_of_target(self, target_location:location.Location, given_location:location.Location) -> "tuple[float, float]":
```
Xierumeng, line 58: This can also be static.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -37,12 +37,72 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
...
+        """
+        Returns the relative coordinates of target w.r.t given location.
+        """
+        relative_location_x = target_location.location_x - given_location.location_x
+        relative_location_y = target_location.location_y - given_location.location_y
+        return (relative_location_x, relative_location_y)
```
Xierumeng, line 64: Remove unneeded parantheses.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -37,12 +37,72 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
...
+        return False
+
+    def next_relative_coordinates_to_target(self,
+                                            target_location:location.Location,
+                                            given_location:location.Location) -> "tuple[float, float]":
+        """
+        Returns the relative x and y coordinates for drone to be sent to.
+        """
+        relative_x, relative_y = self.relative_coordinates_of_target(target_location, given_location)
+        divider = 1
```
Xierumeng, line 83: Remove this since it doesn't do anything.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -37,12 +37,72 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
...
+        else:
+            return (relative_x/divider, relative_y/divider)
+
+    def closest_landing_pad(self,
+                            given_location:location.Location,
+                            landing_pad_locations: "list[location.Location]") -> location.Location:
+        """
+        Finds out the closest landing pad from the given location by checking out their distances.
+        """
+        closest_location = landing_pad_locations[0]
```
Xierumeng, line 97: What if there were no landing pads? (You do not need to change anything, just something to think about).

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -37,12 +37,72 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
...
+        for landing_pad in landing_pad_locations[1:]:
+            distance_from_given_location = DecisionWaypointLandingPads.shortest_distance(landing_pad, given_location)
+            distance_from_closest_location = DecisionWaypointLandingPads.shortest_distance(closest_location, given_location)
```
Xierumeng, line 100: Calculating the distance repeatedly for the closest landing pad is wasteful, is there a way to avoid this?

modules/bootcamp/decision_waypoint_landing_pads.py
Xierumeng: Declare variables (e.g. `self.variable_name` ) in `__init__()` and use those (e.g. `a = self.variable_name` , `self.variable_name = 5` ). Declare as few or as many as you need.

PR #12, opened 2023-09-19
Tong Zhang Bootcamp

TongguangZhang:


modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,33 +87,35 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
         # * conf
         # * device
         # * verbose
-        predictions = ...
+        predictions = self.__model.predict(source=image, device=self.__DEVICE, conf=self.__CONFIDENCE_THRESHOLD)
```
mgupta27, line 90: What about the verbose parameter? Can you disable this parameter?

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,33 +87,35 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
 
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
         # Hint: .shape gets the dimensions of the numpy array
-        # for i in range(0, ...):
+        for i in range(boxes_cpu.shape[0]):
             # Create BoundingBox object and append to list
-            # result, box = ...
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+            bounding_boxes.append(box)
```
mgupta27, line 113: What happens if `result` is false? Do we still want to append to `bounding_boxes`?

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,10 +74,13 @@ def run(self,
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
-        # Do something based on the report and the state of this class...
+        if report.status == drone_status.DroneStatus.HALTED and not self.command_finished:
```
mgupta27, line 77: What would happen if the drone could randomly halt in the middle of a movement? (You do not need to change anything here, just something to think about).

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -37,7 +37,14 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
-        # Add your own
+        self.waypoint_found = False
+        self.landing_pad_found = False
+        self.go_to_waypoint = commands.Command.create_set_relative_destination_command(
+            waypoint.location_x,
+            waypoint.location_y
```
mgupta27, line 44: We encourage bootcampers to follow our [Python Style Guide](https://uwarg-docs.atlassian.net/wiki/spaces/CV/pages/2253226033/Python+Style+Guide). Can you please add a comma after `waypoint.location_y`.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +75,28 @@ def run(self,
...
+            print("Halted at: " + str(report.position))
+            self.landing_pad_found = True
+            min_distance_squared = (60 - (-60)) ** 2 + (60 - (-60)) ** 2
+            closest_pad = 0
+            for landing_pad in landing_pad_locations:
+                distance_squared = (landing_pad.location_x - report.position.location_x) ** 2 \
+                                 + (landing_pad.location_y - report.position.location_y) ** 2
+                if distance_squared < min_distance_squared:
+                    min_distance_squared = distance_squared
+                    closest_pad = landing_pad
```
mgupta27, line 92: `closest_pad` holds a `Location` object here but it is initially set to 0. We want to be consistent with what we store in our variables. Can you please fix this.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +75,28 @@ def run(self,
...
-        # Remove this when done
-        raise NotImplementedError
+        if report.status == drone_status.DroneStatus.HALTED and not self.waypoint_found:
+            print("Halted at: " + str(report.position))
+            self.waypoint_found = True
+            command = self.go_to_waypoint
+        elif report.status == drone_status.DroneStatus.HALTED and not self.landing_pad_found:
+            print("Halted at: " + str(report.position))
+            self.landing_pad_found = True
+            min_distance_squared = (60 - (-60)) ** 2 + (60 - (-60)) ** 2
```
mgupta27, line 85: We don't want the minimum distance to be a finite value. Can we initialize this variable with an infinite value?

modules/bootcamp/decision_waypoint_landing_pads.py
mgupta27: Please fix this in other places in the code as well.

mgupta27 (commented):
Reviewed.

modules/bootcamp/decision_waypoint_landing_pads.py
TongguangZhang: Is it better to have an infinite value or a meaningful value? I should've put a comment, but this is the maximum possible distance within the box - from (-60, -60) to (60, 60)

modules/bootcamp/decision_simple_waypoint.py
TongguangZhang: Could I instead check if the location is the destination specified by the command?

modules/bootcamp/detect_landing_pad.py
```diff
@@ -87,7 +87,12 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
         # * conf
         # * device
         # * verbose
-        predictions = self.__model.predict(source=image, device=self.__DEVICE, conf=self.__CONFIDENCE_THRESHOLD)
+        predictions = self.__model.predict(
+            source=image,
+            device=self.__DEVICE,
+            conf=self.__CONFIDENCE_THRESHOLD,
+            verbose=False
```
mgupta27, line 94: Nit: Can we add a comma after `False`.

modules/bootcamp/detect_landing_pad.py
```diff
@@ -110,7 +115,8 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
         for i in range(boxes_cpu.shape[0]):
             # Create BoundingBox object and append to list
             result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
-            bounding_boxes.append(box)
+            if result:
+                bounding_boxes.append(box)
```
mgupta27, line 119: If a data point is invalid then it makes sense to return no data points from that set rather than skipping over it. The reason why comes from statistics.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -82,8 +82,8 @@ def run(self,
         elif report.status == drone_status.DroneStatus.HALTED and not self.landing_pad_found:
             print("Halted at: " + str(report.position))
             self.landing_pad_found = True
-            min_distance_squared = (60 - (-60)) ** 2 + (60 - (-60)) ** 2
-            closest_pad = 0
+            min_distance_squared = float('inf')
+            closest_pad = location.Location(0, 0)
```
mgupta27, line 86: It might be better to use `None` here rather than `location.Location(0, 0)`. (You do not need to change anything here but just something to think about).

modules/bootcamp/decision_waypoint_landing_pads.py
mgupta27: I would say it is better to have an infinite value since it makes the code more robust. For example, say I make an edit and change the max value to be larger. The code would need to be updated and if it did not get updated we would have a bug in our code. 

It is always better to have a more robust solution since it makes it reduces the chances of having errors later on 😄.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +75,28 @@ def run(self,
...
+            for landing_pad in landing_pad_locations:
+                distance_squared = (landing_pad.location_x - report.position.location_x) ** 2 \
+                                 + (landing_pad.location_y - report.position.location_y) ** 2
+                if distance_squared < min_distance_squared:
+                    min_distance_squared = distance_squared
+                    closest_pad = landing_pad
+            command = commands.Command.create_set_relative_destination_command(
+                closest_pad.location_x - report.position.location_x,
+                closest_pad.location_y - report.position.location_y
```
mgupta27, line 95: Nit, can we add a comma after `location_y` .

mgupta27 (commented):
Reviewed.

modules/bootcamp/decision_waypoint_landing_pads.py
TongguangZhang: makes sense, ty!

mgupta27 (commented):
LGTM!

PR #13, opened 2023-09-20
Leo Tian Bootcamp

leo-cf-tian:


modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,30 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
         # * conf
         # * device
         # * verbose
-        predictions = ...
+        predictions = self.__model(image, conf=0.7, device=self.__DEVICE, verbose=False)
 
         # Get the Result object
-        prediction = ...
+        prediction = predictions[0]             # since only 1 object is returned
```
mgupta27, line 92: Nit: 2 spaces after the code for an inline comment. Please read our [Python Style Guide](https://uwarg-docs.atlassian.net/wiki/spaces/CV/pages/2253226033/Python+Style+Guide) to follow our format and apply them to your bootcamp.

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,30 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
-        boxes_cpu = ...
+        boxes_cpu = boxes_xyxy.detach().cpu().numpy()
 
         # Loop over the boxes list and create a list of bounding boxes
-        bounding_boxes = []
-        # Hint: .shape gets the dimensions of the numpy array
-        # for i in range(0, ...):
-            # Create BoundingBox object and append to list
-            # result, box = ...
+        bounding_boxes = [box[1] for box in [bounding_box.BoundingBox.create(box_cpu) for box_cpu in boxes_cpu] if box[0]]
```
mgupta27, line 107: 1) For tuple assignments we use `ret, val = some_function_call()` . This is specified in our Python Style Guide. This is a common technique we use in our codebase so we want new members to follow it as well 😄.

2) We discourage the use of list initialization within the square brackets as it is very hard to read. Please use the suggested format `for i in range(0, ...):` .

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -67,14 +67,39 @@ def run(self,
...
+        def controlled_destination(p1: location.Location, p2: location.Location):
+            x = p1.location_x - p2.location_x
+            y = p1.location_y - p2.location_y
 
-        # Remove this when done
-        raise NotImplementedError
+            if (abs(x) > 60 or abs(y) > 60):
+                magnitude = max(x, y)
+                x = x / magnitude * 60
+                y = y / magnitude * 60
```
mgupta27, line 82: We don't need to factor in this case as it is guaranteed the waypoint is within the flight boundary so you can remove it. 

Note that if the coordinates did exceed the flight boundary, then we would not want the drone to move at all as the data is invalid.

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -67,14 +67,39 @@ def run(self,
...
+            if (square_dist(report.position, self.waypoint) > self.acceptance_radius**2):
+                x, y = controlled_destination(self.waypoint, report.position)
+
+                command = commands.Command.create_set_relative_destination_command(x, y)
+            else:
+                command = commands.Command.create_land_command()
+
+        elif (report.status == drone_status.DroneStatus.MOVING):
+            if (square_dist(report.position, self.waypoint) > self.acceptance_radius**2):
+                command = commands.Command.create_null_command()
```
mgupta27, line 97: This part is redundant as at the top you will see the command is already initialized to the `null` command. So you can remove this check.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +68,57 @@ def run(self,
...
+        # account for cases where target exceeds flight boundary   
+        def controlled_destination(p1: location.Location, p2: location.Location):
+            x = p1.location_x - p2.location_x
+            y = p1.location_y - p2.location_y
+
+            if (abs(x) > 60 or abs(y) > 60):
+                magnitude = max(x, y)
+                x = x / magnitude * 60
+                y = y / magnitude * 60
+                print(x, y)
```
mgupta27, line 83: Same comment as in task 3.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +68,57 @@ def run(self,
...
+
+                    # redefine waypoint
+                    if (action == "landing_pad"):
+                        self.past_waypoints.append(self.waypoint)
+
+                        # if no landing pads are nearby, return to starting pad
+                        if (len(landing_pad_locations) > 0):
+                            self.waypoint = min(landing_pad_locations, key=lambda location : square_dist(location, report.position))
+                        else:
+                            self.waypoint = self.waypoint = self.past_waypoints[0]
```
mgupta27, line 104: Why do we initialize `self.waypoint` twice here?

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +68,57 @@ def run(self,
...
+
+            # otherwise land
+            else:
+                command = commands.Command.create_land_command()
+
+        elif (report.status == drone_status.DroneStatus.MOVING):
+
+            # if not within distance of waypoint, continue
+            if (square_dist(report.position, self.waypoint) > self.acceptance_radius**2):
+                command = commands.Command.create_null_command()
```
mgupta27, line 117: Same comment as from task 3.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +68,57 @@ def run(self,
...
 
-        # Remove this when done
-        raise NotImplementedError
+            # otherwise land
```
mgupta27, line 119: In the comment you mention the drone lands but you create a halt command here.

mgupta27 (commented):
Reviewed.

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,35 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         bounding_boxes = []
-        # Hint: .shape gets the dimensions of the numpy array
-        # for i in range(0, ...):
-            # Create BoundingBox object and append to list
-            # result, box = ...
+
+        for box_cpu in boxes_cpu:
+            result, box = bounding_box.BoundingBox.create(box_cpu)
+            if (result):
+                bounding_boxes.append(box)
```
mgupta27, line 112: If a data point is invalid (result = False) then it makes sense to return no data points from that set rather than skipping over it. The reason why comes from statistics.

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -67,14 +67,26 @@ def run(self,
         # ============
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
-
-        # Do something based on the report and the state of this class...
-
-        # Remove this when done
-        raise NotImplementedError
+        
+        def square_dist(p1: location.Location, p2: location.Location):
```
mgupta27, line 71: Can you make this a class function rather than a function nested in another function.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -6,7 +6,6 @@
 # Disable for bootcamp use
 # pylint: disable=unused-import
 
-
```
mgupta27, line 9: Do not remove the line here.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +68,41 @@ def run(self,
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
-        # Do something based on the report and the state of this class...
+        def square_dist(p1: location.Location, p2: location.Location):
```
mgupta27, line 71: Same comment as in Task 3.

mgupta27 (commented):
Reviewed.

mgupta27 (commented):
LGTM!

PR #22, opened 2024-01-03
Jane Zeng Autonomy Bootcamp Complete (?)

janez45:


modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -69,9 +75,14 @@ def run(self,
...
-        # Remove this when done
-        raise NotImplementedError
+        if report.status == drone_status.DroneStatus.HALTED and self.command_index < len(self.commands):
+            print("Halted at: "+str(report.position))
+            command = self.commands[self.command_index]
+            self.command_index+=1
+        elif report.status == drone_status.DroneStatus.HALTED and not self.has_sent_landing_command:
+            command = commands.Command.create_land_command()
+            self.has_sent_landing_command = True
+            self.counter+=1
```
mgupta27, line 85: Not sure what `self.counter` is for. If it is used for something can you please reply to this comment to explain what you use it for. If it does not serve any purpose then please remove it.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,9 +99,20 @@ def run(self,
         # ============
 
         # Do something based on the report and the state of this class...
-
-        # Remove this when done
-        raise NotImplementedError
+        if report.status == drone_status.DroneStatus.HALTED and self.command_index < len(self.commands):
+            print("Halted at: "+str(report.position))
+            command = self.commands[self.command_index]
+            self.command_index+=1
```
mgupta27, line 105: You don't have to change anything but it is something to think about. If `self.commands` only holds one element then it doesn't make a lot of sense to keep it as an array. Instead you could store it in a variable and use a boolean flag variable to indicate when you have arrived at the waypoint.

mgupta27 (commented):
Reviewed.

Overall good start. Just a couple of comments and nits that you need to address. I highly suggest you read our style guide and adhere to it as you will need to follow it when you finish the bootcamp and start taking on tasks. [WARG Python Style Guide](https://uwarg-docs.atlassian.net/wiki/x/MYBNhg).

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,39 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+
         # Hint: .shape gets the dimensions of the numpy array
         # for i in range(0, ...):
             # Create BoundingBox object and append to list
             # result, box = ...
-
+        
+        for bbox in boxes_cpu:
+            result, new_bounding_box = bounding_box.BoundingBox.create(bbox)
+            if result:
```
mgupta27, line 116: If the result is `False` then we don't want to include an image that has an invalid data point. Instead, can you change the loop so that it returns either all of the bounding boxes in the image or no bounding boxes in the image.

mgupta27 (commented):
Reviewed. 

Next time please do not resolve comments from the reviewer. The reviewer will resolve comments once the issue has been fixed. From the reviewers point of view, comments are a good way to go back and remind themselves what the previous issues were and they can deem if the issue has been resolved.

mgupta27 (commented):
LGTM.

PR #23, opened 2024-01-09
Nathan - Bootcamp

nsmartinx:


modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,10 +70,25 @@ def run(self,
...
+            if self.distance_x > 0.1:
+                command = commands.Command.create_set_relative_destination_command(min(60, self.distance_x),  0)
+            elif self.distance_x < -0.1:
+                command = commands.Command.create_set_relative_destination_command(max(-60, self.distance_x),  0)
+            elif self.distance_y > 0.1:
+                command = commands.Command.create_set_relative_destination_command(0, min(60, self.distance_y))
+            elif self.distance_y < -0.1:
+                command = commands.Command.create_set_relative_destination_command(0, max(-60, self.distance_y))
+            else:
+                command = commands.Command.create_land_command()
```
mgupta27, line 86: A few comments.

* Currently the way you are travelling the map is by first moving in the x-direction and then the y-direction. This is not ideal as the drone is traveling a longer distance to reach the waypoint (this means more energy is used during its flight which isn't good as our competition drone has a limited amout of power). Instead, it would be more efficient if we could travel diagonally. Think about it in terms of a right-angled triangle. It would be less distance to travel across the hypotenuse rather than the lengths of the other two sides.
* Rather than checking if `self.distance_x` is greater than 0.1 or less than -0.1 you can simplify this by using the absolute value. e.g. `abs(self.distance_x) > 0.1`.
  * Could also be a good idea to have the value of 0.1 stored as a constant variable.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -17,6 +17,8 @@
 # Disable for bootcamp use
 # pylint: disable=unused-argument,line-too-long
 
+def pad_distance_squared(pad, drone):
```
mgupta27, line 20: You don't have to change anything here. On our team we like to specify the typing of parameters in a function and also the return type. It helps to make our code more readable and easier for other team members to understand what is going on.

modules/bootcamp/decision_simple_waypoint.py
mgupta27: The same comment is also applied to Task 4 as well.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,9 +78,32 @@ def run(self,
...
+            elif self.distance_y > 0.1:
+                command = commands.Command.create_set_relative_destination_command(0, min(60, self.distance_y))
+            elif self.distance_y < -0.1:
+                command = commands.Command.create_set_relative_destination_command(0, max(-60, self.distance_y))
+            elif self.reached_waypoint:   
+                command = commands.Command.create_land_command()
+            else:
+                self.reached_waypoint = True
+                closest_pad = landing_pad_locations[0]
+                closest_pad_distance_squared = 999999
```
mgupta27, line 98: Rather than specifying a value of `999999` it might make more sense to set it to infinity rather than an arbitrarily large number. It will be more evident that you are taking the maximum possible value to start of with.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,9 +78,32 @@ def run(self,
...
+                for pad in landing_pad_locations:
```
mgupta27, line 99: Instead of `pad`, can you change the variable name to `landing_pad_location`. The reason is because we want to follow our Python Style guide. For iterating through a list you can see the link [here](https://uwarg-docs.atlassian.net/wiki/spaces/CV/pages/2253226033/Python+Style+Convention#Pythonic).

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,40 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
         # * conf
         # * device
         # * verbose
-        predictions = ...
+
```
mgupta27, line 89: Can you remove the extra space above here. Please refer to our style guide and fix any other styling issues in your code. [Python Style Guide](https://uwarg-docs.atlassian.net/wiki/spaces/CV/pages/2253226033/Python+Style+Convention)

mgupta27: Also remove extra spaces in the other places in this file.

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -13,6 +13,14 @@
 from .. import location
 from ..private.decision import base_decision
 
+def clamp(number: float,
+          min: float,
+          max: float) -> float:
+    if number < min:
```
mgupta27, line 19: I think it would be better to have `min` and `max` has class constants. That way you don't have to keep passing in min and max. Also, I think you should make this a class function.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -17,6 +17,19 @@
...
+        return (pad.location_x - drone.location_x) ** 2 + (pad.location_y - drone.location_y) ** 2
+
+def clamp(number: float,
+          min: float,
+          max: float) -> float:
+    if number < min:
+        return min
+    if number > max:
+        return max
+    return number
```
mgupta27, line 32: Move these functions to class functions.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +86,25 @@ def run(self,
...
+                if self.reached_waypoint:
+                    command = commands.Command.create_land_command()
+                else:
+                    self.reached_waypoint = True
+                    closest_pad = landing_pad_locations[0]
+                    closest_pad_distance_squared = float('inf')
+                    for landing_pad_location in landing_pad_locations:
+                        if pad_distance_squared(landing_pad_location, report.position) < closest_pad_distance_squared:
+                            closest_pad_distance_squared = pad_distance_squared(landing_pad_location, report.position)
+                            closest_pad = landing_pad_location
```
mgupta27, line 103: Consider breaking this into multiple lines (you can append multiple lines with a backslash).

modules/bootcamp/detect_landing_pad.py
```diff
@@ -11,7 +11,7 @@
 
 from .. import bounding_box
 
-
+
```
mgupta27, line 14: Remove extra space here.

modules/bootcamp/decision_waypoint_landing_pads.py
mgupta27, line 32: You don't have to change anything here but its something to think about. 

In our Autonomy codebase, we do extensive error checking to ensure we are getting the correct data. If we find that the data we receive is incorrect, rather than trying to correct it we will return a `False, None` response indicating faulty data was provided.

The rationale behind this is that we don't want to act on faulty data. Instead, we want to disregard it and log it.

mgupta27 (commented):
Reviewed.

mgupta27 (commented):
LGTM!

PR #25, opened 2024-01-10
W24 Bootcamp Completed - Mahan

MahanSharifi:


modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,33 +86,38 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Hint: .shape gets the dimensions of the numpy array
         # for i in range(0, ...):
             # Create BoundingBox object and append to list
             # result, box = ...
-
-        # Remove this when done
-        raise NotImplementedError
-
+        for i in boxes_cpu:
+            result, box = bounding_box.BoundingBox.create(np.array(i))
```
Trotyl15, line 113: Improve readability here. BoundingBox should be created using items from boxes_cpu which is `i` as you defined. But `i` usually refer to the index number so rename it to avoid potential confusion. Also, doing `np.array()` is unnessesary.

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -69,9 +70,18 @@ def run(self,
...
-
-        # Remove this when done
-        raise NotImplementedError
+        position = report.position
+        status = report.status
+        rel_x_dest = self.waypoint.location_x - position.location_x
+        rel_y_dest = self.waypoint.location_y - position.location_y
+        halted = drone_status.DroneStatus.HALTED
+
+        if status == halted:
```
Trotyl15, line 79: What if the drone halts for some unrelated reason while on the way to the waypoint? (You do not need to change anything here, just something to think about).

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +70,34 @@ def run(self,
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
+        def distance_between_locations(location_one: location.Location, location_two: location.Location):
+            return ((location_one.location_x - location_two.location_x)**2 + (location_one.location_y - location_two.location_y)**2)**0.5
```
Trotyl15, line 74: The square root operation is computationally expensive, can you think of a better way to avoid it? (You don't need to change anything here, just something to think about)

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +70,34 @@ def run(self,
...
+        
+        status = report.status
+        halted = drone_status.DroneStatus.HALTED
         # Do something based on the report and the state of this class...
 
-        # Remove this when done
-        raise NotImplementedError
+        if status == halted:
+            # Case 1: moving to the waypoint
```
Trotyl15, line 81: Good use of commenting! (You don't need to change anything here)

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +70,34 @@ def run(self,
...
+            if not self.away_from_landing_pad and not self.away_from_waypoint:
+                command = commands.Command.create_set_relative_destination_command(self.waypoint.location_x - report.position.location_x, self.waypoint.location_y - report.position.location_y)
+                self.away_from_landing_pad = True
+
+            # Case 2: moving to the nearest landing pad only when at waypoint
+            elif not self.away_from_waypoint and self.away_from_landing_pad:
+                nearest_pad = landing_pad_locations[0]
```
Trotyl15, line 88: While the bootcamp specifies at least 1 landing pad, what happens if there were no landing pads? (You do not need to change anything here, just something to think about).

modules/bootcamp/detect_landing_pad.py
MahanSharifi: Makes sense. I made the change.

modules/bootcamp/decision_simple_waypoint.py
MahanSharifi: For one we could implement drone_status.DroneStatus.MOVING for better routing, but to answer your question, we should probably have a diagnostic routine handler for unexpected halting.

modules/bootcamp/decision_waypoint_landing_pads.py
MahanSharifi: Just compare the squared distances.

modules/bootcamp/decision_waypoint_landing_pads.py
MahanSharifi: Check if landing_pad list is not empty, check to see if landing_pad is available to land, if not possible to land, take another route.

modules/bootcamp/decision_simple_waypoint.py
Trotyl15: Yes, and we can also add a checker to see if the difference in coordinate of landing pad and drone is less than 0.1. If it is than land, if not set destination again or handle the expection.

MahanSharifi: Makes sense. Is 0.1 standard?

Trotyl15 (approved):
LGTM!

modules/bootcamp/decision_simple_waypoint.py
Trotyl15: No, but we specified 0.1 as the maximum error in distance in the bootcamp instruction for this particular simulator.

PR #26, opened 2024-01-15
Bootcamp - Hard Shah

hardshah:


modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,33 +86,34 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
-        # for i in range(0, ...):
-            # Create BoundingBox object and append to list
-            # result, box = ...
-
-        # Remove this when done
-        raise NotImplementedError
+        for i in range(0, len(boxes_cpu)):
+            #Create BoundingBox object and append to list
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+            if result:
```
maxlou05, line 113: It is generally unideal to keep the other results if one of the boxes is invalid. It is preferred to scrap the whole data point (in this case, the whole image). (You do not need to change anything here, just something to think about)

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -36,8 +36,11 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
         # ============
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
-
-        # Add your own
+        self.command_index = 0
+        self.commands = [
+            commands.Command.create_set_relative_destination_command( waypoint.location_x,  waypoint.location_y)
```
maxlou05, line 41: Note that this is a relative destination command. Just using the waypoint's absolute location would result in moving to an incorrect position.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -36,12 +34,33 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
         # ============
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
+        self.command_index = 0
+        self.commands = [
+            commands.Command.create_set_relative_destination_command( waypoint.location_x,  waypoint.location_y)
```
maxlou05, line 39: Same as before, this is a relative position command.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -36,12 +34,33 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
...
+            squared_distance = self.calculate_squared_distance(pad.location_x, pad.location_y, current.location_x, current.location_y)
+            if squared_distance < smallest_squared_distance:
+                smallest_squared_distance, closest_pad = squared_distance, pad
+        
+        return closest_pad
+
+    def calculate_squared_distance(self, x1, y1, x2, y2):
+        dx = x2-x1
+        dy = y2-y1
+        return ((dx*dx)+(dy*dy))
```
maxlou05, line 63: Just a note, you can use dx**2 in python to do power. (You do not need to change anything here).

modules/bootcamp/detect_landing_pad.py
hardshah: Ok, so I guess I can add an else statement that can remove all the bounding boxes, if the box is invalid?

modules/bootcamp/decision_waypoint_landing_pads.py
hardshah: Okay, that sounds good, I tried avoiding it because I read somewhere that multiplication is less expensive than exponentiation in python.

modules/bootcamp/decision_simple_waypoint.py
hardshah: I'm unsure why the absolute location would be incorrect. If the drone is starting at the origin, then we can use the absolute location of the waypoint as well right?

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -69,12 +72,16 @@ def run(self,
...
 
         # Do something based on the report and the state of this class...
-
-        # Remove this when done
-        raise NotImplementedError
+        if report.status == drone_status.DroneStatus.HALTED and self.command_index < len(self.commands): #Fly to waypoint
+            command = self.commands[self.command_index]
+            self.command_index+=1
+        
+        elif report.status == drone_status.DroneStatus.HALTED and not self.has_sent_landing_command: # Land
```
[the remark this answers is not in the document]
hardshah: Maybe we can check the location, and see if it completed the movement, and if not we can send the command again?

modules/bootcamp/decision_waypoint_landing_pads.py
maxlou05: Oh, I see. Python is slow anyway, I don't think such a little change will do much. I think it's more geared toward readability.

modules/bootcamp/decision_simple_waypoint.py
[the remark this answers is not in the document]
maxlou05: Nice! Good idea.

modules/bootcamp/detect_landing_pad.py
maxlou05: Yes, that would work.

modules/bootcamp/decision_simple_waypoint.py
maxlou05: I suppose, but generally, it's better to be safe. Additionally, it makes more sense to readers.

maxlou05: You can just make another commit to fix this, and then you should be good to go.

hardshah: Ok should be fixed

maxlou05:
LGTM!

PR #28, opened 2024-01-16
Autonomy Bootcamp - Ohm Patel

Ohm-Patel:


modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,32 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
 
         # Get the xyxy boxes list from the Boxes object in the Result object
-        boxes_xyxy = ...
+        boxes_xyxy = prediction.boxes.xyxy
 
         # Detach the xyxy boxes to make a copy,
         # move the copy into CPU space,
         # and convert to a numpy array
-        boxes_cpu = ...
+        boxes_cpu = boxes_xyxy.cpu().numpy()
```
mgupta27, line 104: In this line you are not detaching the `xyxy` box.

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,32 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
 
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
         # Hint: .shape gets the dimensions of the numpy array
-        # for i in range(0, ...):
+        for i in range(0, boxes_cpu.shape[0]):
             # Create BoundingBox object and append to list
-            # result, box = ...
+            result, box = bounding_box.BoundingBox.create(np.array([boxes_cpu[i, 0], boxes_cpu[i, 1], boxes_cpu[i, 2], boxes_cpu[i, 3]]))
```
mgupta27, line 111: `boxes_cpu` is already a numpy array so you can pass it in directly. No need to index it and create a new numpy array.

mgupta27, line 111: What happens if `result = False`. We don't want to append the bounding box if that occurs.

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -69,9 +73,16 @@ def run(self,
         # ============
 
         # Do something based on the report and the state of this class...
+        distance = (self.waypoint.location_x - report.position.location_x)**2 + (self.waypoint.location_y - report.position.location_y)**2
 
-        # Remove this when done
-        raise NotImplementedError
+        if drone_status.DroneStatus.HALTED == report.status:
+            if self.flag:
```
mgupta27, line 79: Nit: Come up with a more specific name for `self.flag`. Currently it is very vague.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,9 +73,36 @@ def run(self,
         # ============
 
         # Do something based on the report and the state of this class...
+        def find_distance(location_one, location_two):
+            return (location_two.location_x - location_one.location_x)**2 + (location_two.location_y - location_one.location_y)**2
+        
+        def find_closest_landing_pad(curr_location, landing_pad_list):
+            closest_landing_pad = None
+            closest_distance = 10000000000000
```
mgupta27, line 81: Rather than setting `closest_distance` to an arbitrarily large number, it makes more sense to set it to infinity. That way its clear you have set this variable to the largest possible value.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,9 +73,36 @@ def run(self,
...
+            for index in landing_pad_list:
```
mgupta27, line 82: Nit: Find a better name for `index` as it very broad. Perhaps `landing_pad`.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,9 +73,36 @@ def run(self,
         # ============
 
         # Do something based on the report and the state of this class...
+        def find_distance(location_one, location_two):
```
mgupta27, line 76: Instead of having these functions within the `run` function, can you make them class functions.

modules/bootcamp/decision_simple_waypoint.py
mgupta27: Same comment in task 4 code.

mgupta27 (commented):
Reviewed.

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,35 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
         # Hint: .shape gets the dimensions of the numpy array
-        # for i in range(0, ...):
+        for i in boxes_cpu:
             # Create BoundingBox object and append to list
-            # result, box = ...
+            result, box = bounding_box.BoundingBox.create(i)
+            if not result:
+                bounding_boxes = []
```
mgupta27, line 113: If the result is False then you set `bounding_boxes` to be an empty list, but you still loop through the for loop. You need to break out of the loop.

mgupta27 (commented):
Reviewed.

mgupta27 (commented):
LGTM

PR #30, opened 2024-01-19
Khushi Bootcamp

KhushiJirge:


modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -70,8 +73,18 @@ def run(self,
 
         # Do something based on the report and the state of this class...
 
+        
+        if report.status == drone_status.DroneStatus.HALTED and self.is_at_waypoint == False:
+            command = commands.Command.create_set_relative_destination_command((self.waypoint.location_x - report.position.location_x), \
+                                                                               (self. waypoint.location_y - report.position.location_y))
```
maxlou05, line 79: No need for brackets and backslash

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -70,8 +73,18 @@ def run(self,
...
+            self.is_at_waypoint = True
+
+        elif report.status == drone_status.DroneStatus.HALTED and self.should_land == False:
```
maxlou05, line 82: While the bootcamp specifies that the drone will always complete its movement, what if this wasn't guaranteed and it could halt on its own at any random time? (You do not need to change anything here, just something to think about).

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -70,8 +73,18 @@ def run(self,
...
+            command = commands.Command.create_land_command()
+            self.should_land = True
+
         # Remove this when done
-        raise NotImplementedError
+        # raise NotImplementedError
```
maxlou05, line 87: You can delete the line, no need to keep it as a comment (it's not a very helpful comment, and if you need it back git is there).

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -70,11 +73,46 @@ def run(self,
 
         # Do something based on the report and the state of this class...
 
+        def find_shortest_distance(curr_location, landing_pad_locations_list):
+            max_distance = float('inf')
```
maxlou05, line 77: the variable name is max_distance while it records the minimum distance. This is kind of confusing for people reading it, so consider changing the name.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -70,11 +73,46 @@ def run(self,
...
+            closest_waypoint = landing_pad_locations_list[0]
+            for landing_pad in landing_pad_locations_list:
+                x = landing_pad.location_x - curr_location.location_x
+                y = landing_pad.location_y - curr_location.location_y
+
+                distance = ((x ** 2)+(y ** 2)) ** 1/2
```
maxlou05, line 83: Computing the square root is expensive. Is there a way to find the closest waypoint without this? (also, it will do ((x^2 + y^2 ) ^ 1) /2 because of operator precedence)

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -70,11 +73,46 @@ def run(self,
...
+                if distance < max_distance:
+                    max_distance = distance
+                    closest_waypoint = landing_pad
+
+            return closest_waypoint
+
+        if report.status == drone_status.DroneStatus.HALTED and self.is_at_waypoint == False:
+            command = commands.Command.create_set_relative_destination_command((self.waypoint.location_x - report.position.location_x), \
+                                                                               (self. waypoint.location_y - report.position.location_y))
```
maxlou05, line 92: Backslash and brackets are unnecessary

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -70,11 +73,46 @@ def run(self,
...
+
+        elif report.status == drone_status.DroneStatus.HALTED and self.is_at_closest_waypoint == True:
+            command = commands.Command.create_land_command()
+
+        
+        
+
         # Remove this when done
-        raise NotImplementedError
+        # raise NotImplementedError
```
maxlou05, line 110: Again, no need to leave this as a comment, you can remove it.

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,41 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
         # * conf
         # * device
         # * verbose
-        predictions = ...
+        predictions = self.__model.predict(source=image, conf=0.7, device=self.__DEVICE, verbose=False)
 
         # Get the Result object
-        prediction = ...
+        prediction = predictions.__getitem__(0)
```
maxlou05, line 92: If this works, that's fine. But you can just do predictions[0] as well. (You do not need to change anything here, just something to think about)

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,41 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Get the xyxy boxes list from the Boxes object in the Result object
-        boxes_xyxy = ...
+        boxes = prediction.boxes.cpu().numpy()
+        boxes_xyxy = boxes.xyxy
 
         # Detach the xyxy boxes to make a copy,
         # move the copy into CPU space,
         # and convert to a numpy array
-        boxes_cpu = ...
+        boxes_cpu = boxes_xyxy.copy()
```
maxlou05, line 106: I guess this works, but now you have 2 variables pointing to the same thing. Also, you have copied the entire boxes from GPU to CPU. I think the intention was just to have the boxes.xyxy copied down to CPU, not the whole boxes list.

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,41 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
-        # for i in range(0, ...):
+
+        #.shape returns a tuple with each index having corresponding elements
+        for i in range(0, boxes_cpu.shape[0]):
             # Create BoundingBox object and append to list
-            # result, box = ...
+            success, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+
+            if box != None or success == True:
+                bounding_boxes.append(box)
```
maxlou05, line 118: Generally, in statistics, if one part of the dataset is invalid, we do not use the rest of the dataset. So if one of the boxes were invalid, we get rid of the entire image. (You do not have to change anything here, just something to think about)

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,41 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+        
+        return(bounding_boxes, image_annotated)
```
maxlou05, line 120: Brackets are unnecessary

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,41 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
 
         # Remove this when done
-        raise NotImplementedError
+        #raise NotImplementedError
```
maxlou05, line 123: Same here, no need to keep this as a comment

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -70,11 +73,40 @@ def run(self,
...
-        # Remove this when done
-        raise NotImplementedError
+        def find_shortest_distance(curr_location, landing_pad_locations_list):
+            min_distance = float('inf')
+            closest_waypoint = landing_pad_locations_list[0]
+            for landing_pad in landing_pad_locations_list:
+                x = landing_pad.location_x - curr_location.location_x
+                y = landing_pad.location_y - curr_location.location_y
+
+                distance = ((x ** 2)+(y ** 2)) ** (1/2)
```
maxlou05, line 83: Although this is fine, try to think of another way to determine the closest landing pad without using a square root calculation for optimization (computing square roots is slow).

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,38 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Plot the annotated image from the Result object
         # Include the confidence value
-        image_annotated = ...
+        image_annotated = prediction.plot(conf=True)
+
 
         # Get the xyxy boxes list from the Boxes object in the Result object
-        boxes_xyxy = ...
+        boxes = prediction.boxes.cpu().numpy()
+        boxes_xyxy = boxes.xyxy
```
maxlou05, line 101: You are still copying the entire Boxes object into CPU here. I think the intention is to get a reference to just the xyxy boxes list, and then copy it to the CPU later in the next step.

maxlou05 (approved):
LGTM!

maxlou05:
LGTM!

PR #32, opened 2024-01-26
Kenson Hui PR

kensonhui:


modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,33 +86,36 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+        prediction = predictions[0]
+
+        # No prediction -> return original image
+        if not prediction:
+            return ([], image)
 
         # Plot the annotated image from the Result object
         # Include the confidence value
-        image_annotated = ...
+        image_annotated = prediction.plot()
```
maxlou05, line 99: Please include the confidence value in the image.

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,33 +86,36 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
-        # Hint: .shape gets the dimensions of the numpy array
-        # for i in range(0, ...):
+        for i in range(0, boxes_cpu.shape[0]):
             # Create BoundingBox object and append to list
-            # result, box = ...
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+            if result:
+                bounding_boxes.append(box)
```
maxlou05, line 115: In statistics, an invalid portion of the data point involves removing the entirety of the data point (where the data point here is an image, as each image is independent). It's ok for now though (you do not need to change anything, just something to think about).

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -69,12 +69,26 @@ def run(self,
...
-        raise NotImplementedError
+        if report.status == drone_status.DroneStatus.HALTED and not self.has_sent_move_command:
+            # Calculate relative destination
+            relative_x, relative_y = self.calculate_relative_distance(waypoint=self.waypoint, 
+                                                                      drone_location=report.position)
+            command = commands.Command.create_set_relative_destination_command(relative_x, relative_y)
+            print("Drone location:", report.position)
+            print("Travelling to", self.waypoint)
+            print("Relative Distance is ", relative_x, relative_y)
+            self.has_sent_move_command = True
```
maxlou05, line 81: This is good, it gets the drone to the desired waypoint. However, we also want the drone to land, and not hover around in the air at the end. Just add code that lands the drone once it reaches the end.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,13 +69,60 @@ def run(self,
...
+        return distance_x ** 2 + distance_y ** 2
+    
+    def find_closest_landing_pad(self, 
+                                 waypoint: location.Location, 
+                                 landing_pad_locations: "list[location.Location]") -> location.Location:
+        if len(landing_pad_locations) == 0:
+            raise ValueError
+
+        closest_landing_pad = min(landing_pad_locations, 
+                                  key=lambda location: self.calculate_euclidian_distance(location, self.waypoint))
```
maxlou05, line 126: Nice use of the built-in min and lambda functions! (You do not need to change anything here)

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,13 +69,60 @@ def run(self,
...
+        Given two locations, return the euclidian distance of the two
+        """
+        distance_x, distance_y = self.calculate_relative_difference(destination, starting_point)
+        return distance_x ** 2 + distance_y ** 2
+    
+    def find_closest_landing_pad(self, 
+                                 waypoint: location.Location, 
+                                 landing_pad_locations: "list[location.Location]") -> location.Location:
+        if len(landing_pad_locations) == 0:
+            raise ValueError
```
maxlou05, line 123: I believe that you do not want to raise errors, as those will crash the program. In the case that your drone is in the air and the program crashes, it might fall and break. However, the bootcamp never specified what you should do when there are no landing pads, so it's fine for now. (You do not need to change anything, just something to think about)

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,13 +69,60 @@ def run(self,
...
+    
+    def calculate_euclidian_distance(self,
+                                     destination: location.Location,
+                                     starting_point: location.Location
+                                     ) -> float:
+        """
+        Given two locations, return the euclidian distance of the two
+        """
+        distance_x, distance_y = self.calculate_relative_difference(destination, starting_point)
+        return distance_x ** 2 + distance_y ** 2
```
maxlou05, line 117: Good use of distance squared, as that is sufficient to prove that the distance is greater or lower in comparisons. However, please do name your function and doc string to reflect that it is not actually the euclidean distance, but the distance squared.

maxlou05 (commented):
Reviewed

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,13 +69,60 @@ def run(self,
...
 
         return command
+
+
+    def calculate_relative_difference(self, 
+                                    destination: location.Location, 
+                                    starting_point: location.Location) -> (float, float):
+        """
+        Given the absolute location of the waypoint and the drone, return the relative distance of the two
+        """
```
maxlou05, line 105: Great use of docstrings and type hinting in your functions! (You do not need to change anything here)

modules/bootcamp/detect_landing_pad.py
kensonhui: I noticed in the ultralytics docs that conf=True by default and I have made the changes for better clarity

modules/bootcamp/decision_simple_waypoint.py
kensonhui: You're right fixed now.

modules/bootcamp/decision_waypoint_landing_pads.py
kensonhui: You're right I forgot to change the name after I changed it to not use the square root, fixed now

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,33 +86,36 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
         # * conf
         # * device
         # * verbose
-        predictions = ...
-
+        predictions = self.__model.predict(source=image)
```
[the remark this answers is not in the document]
kensonhui: I found an interesting bug! I'll add it to the discord thread

maxlou05 (approved):
Approved.

PR #35, opened 2024-01-31
Victor's PR

deltag0:


modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -18,16 +18,30 @@
...
+    
+
 # All logic around the run() method
 # pylint: disable-next=too-few-public-methods
 class DecisionWaypointLandingPads(base_decision.BaseDecision):
     """
     Travel to the designed waypoint and then land at the nearest landing pad.
     """
+    
+
```
TongguangZhang, line 32: No spaces between docstring and code. Also, make sure you aren't editing anything outside of the comment guards (the parts that say bootcampers modify above/below).

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -18,16 +18,30 @@
...
         Initialize all persistent variables here with self.
         """
+        self.visited = False
+        # self.l = []
+        # self.landing_count = int(input("Enter how many landing locations you want: "))
+        # for x in range(self.landing_count):
+        #     self.ix = float(input("Enter the x coordinate for the landing location: "))
+        #     self.iy = float(input("Enter the y coordinate for the landing location: "))
+        #     self.l.append((self.ix, self.iy))
+        # print(self.l)
```
TongguangZhang, line 44: Try not to have commented out or unreachable code - we use Git to keep track of changes so we can always go back in the revision history if we need this code again. Also, make sure you're not editing anything outside of the comment guards.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -47,7 +61,7 @@ def run(self,
             report: drone_report.DroneReport,
             landing_pad_locations: "list[location.Location]") -> commands.Command:
         """
-        Make the drone fly to the waypoint and then land at the nearest landing pad.
+        Make the drone fly to the waypoint.
```
TongguangZhang, line 64: Make sure you're not editing anything outside of the comment guards.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -61,17 +75,77 @@ def run(self,
...
+            mins = -1
+            i = None
+            for x in range(len(l.copy())):
+                dis = distance_calculator(startx, starty, l[x].location_x, l[x].location_y)
+                if mins < 0:
+                    mins = dis
+                    i = x
+                elif dis < mins:
+                    mins = dis
+            return i
```
TongguangZhang, line 91: Make sure you're not editing anything outside of the comment guards. Don't put these functions inside the run method - we don't want to declare them every time run is called. Since these are helper functions and not class functions, place them outside of the class

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -61,17 +75,77 @@ def run(self,
...
+            #     i = len(l) - 1
+            #     l.pop()
+            #     return find_min(l, dis, startx, starty, i)
+            # l.pop()
+            # return find_min(l, mins, startx, starty, i)
         # Default command
+       
         command = commands.Command.create_null_command()
-
+
```
TongguangZhang, line 113: Remove commented out code (see above). Make sure you're not editing anything outside of the comment guards, and avoid training whitespace (line 113).

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -61,17 +75,77 @@ def run(self,
...
-
+        
+        pos = report.position
         # Do something based on the report and the state of this class...
+        print(report.status)
+        
+        
+
+        # if report.status == drone_status.DroneStatus.HALTED and len(self.l) != 0:
+        #     command = commands.Command.create_set_relative_destination_command(self.l[-1][0] - pos.location_x, self.l[-1][1] - pos.location_y)
```
TongguangZhang, line 125: Remove commented out code (see above).

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -61,17 +75,77 @@ def run(self,
...
+
+        if pos == location.Location(0.0,0.0):
+            print(self.waypoint)
+            command = commands.Command.create_set_relative_destination_command(self.waypoint.location_x, self.waypoint.location_y)
```
TongguangZhang, line 129: Consider what happens if the drone doesn't start at the centre, how else can we check if the drone hasn't started moving yet? This also means you would have to use the relative distance instead of absolute.

modules/bootcamp/decision_waypoint_landing_pads.py
TongguangZhang, line 91: Try to make this more pythonic - we prefer using
```
for thing in things:
```
and return the landing pad object instead of
```
for i in range(len(things)):
```

also - there's no need to make a copy of the list.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -61,17 +75,77 @@ def run(self,
...
+        if self.visited == True and report.status == drone_status.DroneStatus.HALTED:
+            command = commands.Command.create_land_command()
+        if report.status == drone_status.DroneStatus.HALTED and pos == self.waypoint:
+            close = int(find_min(landing_pad_locations, pos.location_x, pos.location_y))
+            print("HELLO", close)
+            print(type(close))
```
TongguangZhang, line 135: Remove these debugging print statements.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -61,17 +75,77 @@ def run(self,
...
+        
+
+        # if report.status == drone_status.DroneStatus.HALTED and len(self.l) != 0:
+        #     command = commands.Command.create_set_relative_destination_command(self.l[-1][0] - pos.location_x, self.l[-1][1] - pos.location_y)
+
+        if pos == location.Location(0.0,0.0):
+            print(self.waypoint)
+            command = commands.Command.create_set_relative_destination_command(self.waypoint.location_x, self.waypoint.location_y)
+        if self.visited == True and report.status == drone_status.DroneStatus.HALTED:
+            command = commands.Command.create_land_command()
```
TongguangZhang, line 131: Make sure you check that the drone is at the landing pad before landing (use the acceptance radius and the location of the landing pad that you found).

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -61,17 +75,77 @@ def run(self,
...
+        if report.status == drone_status.DroneStatus.HALTED and pos == self.waypoint:
+            close = int(find_min(landing_pad_locations, pos.location_x, pos.location_y))
+            print("HELLO", close)
+            print(type(close))
+            self.visited = True
+            command = commands.Command.create_set_relative_destination_command(landing_pad_locations[close].location_x - pos.location_x, landing_pad_locations[close].location_y - pos.location_y)
```
TongguangZhang, line 137: Make sure you don't have lines that are too long (over 100 characters). If you're using a code editor like vscode, it has a ruler utility that will show you where the 100 character line is.

modules/bootcamp/detect_landing_pad.py
```diff
@@ -74,6 +79,13 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
         Return: A tuple of (list of bounding boxes, annotated image) .
             The list of bounding boxes can be empty.
         """
+        def flattener(l):
+         return_list = []
+         for x in l:
+            if x.__class__ == list:
+                for y in x:
+                    return_list.append(y)
+         return return_list
```
TongguangZhang, line 88: Make sure you're not defining functions you don't use, or edit anything outside the comment guards.

modules/bootcamp/detect_landing_pad.py
```diff
@@ -12,6 +12,11 @@
 from .. import bounding_box
 
 
+            
+
+
+
+
```
TongguangZhang, line 19: Only 2 spaces between these lines (between the import statements and comment). Also, make sure you aren't editing anything outside of the comment guards (the parts that say bootcampers modify above/below).

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +98,43 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
         # * conf
         # * device
         # * verbose
-        predictions = ...
-
+        predictions = self.__model.predict(image, conf = 0.7, device = DetectLandingPad.__DEVICE)
+
```
TongguangZhang, line 102: Training whitespace.

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +98,43 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
         # * conf
         # * device
         # * verbose
-        predictions = ...
-
+        predictions = self.__model.predict(image, conf = 0.7, device = DetectLandingPad.__DEVICE)
```
TongguangZhang, line 101: 1. Make sure you call self for the class' members instead of the class name.
2. Try disabling the verbose parameter
3. Set a class member for the confidence threshold - we want to avoid "magic numbers"

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +98,43 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Get the Result object
-        prediction = ...
-
+        prediction =  predictions[0]
+        print(predictions)
+        
         # Plot the annotated image from the Result object
         # Include the confidence value
-        image_annotated = ...
+        image_annotated =  prediction.orig_img
```
TongguangZhang, line 109: We're looking for the resulting image here with the annotations, not the original without the detections. Try looking in the Ultralytics documentation for plotting the detection results.

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +98,43 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+       
+        print(boxes_cpu)
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
+        for box in boxes_cpu:
+            print(box)
+            adding_box = bounding_box.BoundingBox.create(box)[1]
+            print(adding_box)
+            bounding_boxes.append(adding_box)
+        print(bounding_boxes)
```
TongguangZhang, line 129: Let's remove some of the print statements here to clean up the code.

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +98,43 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+        boxes_cpu = boxes_xyxy.detach().cpu().numpy()
+       
+        print(boxes_cpu)
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
+        for box in boxes_cpu:
+            print(box)
+            adding_box = bounding_box.BoundingBox.create(box)[1]
+            print(adding_box)
+            bounding_boxes.append(adding_box)
```
TongguangZhang, line 128: What happens if the create function fails, do we want to add a null value to the list?

<details>
  <summary>Answer: </summary>
  We want to return an empty list (nothing)
</details>

TongguangZhang (commented):
Great job Victor. Most of the logic is sound, but there are a couple small flaws I've noticed.

Other than that, take a look at our python style guide.
https://uwarg-docs.atlassian.net/wiki/spaces/CV/pages/2253226033/Python+Style+Convention#Going-against-convention

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -61,18 +62,25 @@ def run(self,
...
+        w_x = self.waypoint.location_x
+        w_y = self.waypoint.location_y
+        r = self.acceptance_radius
+        halted = drone_status.DroneStatus.HALTED
+        move = commands.Command.create_set_relative_destination_command
+        pos = report.position
+        if report.status == halted and self.visited == False:
+            print(self.waypoint)
+            command = move(w_x - pos.location_x, w_y - pos.location_y)
+        if report.status == halted and (w_x - pos.location_x)**2 + (w_y - pos.location_y)**2 <= r**2:
```
TongguangZhang, line 82: we usually specify that it's less than the radius, not less than or equal to, because the equals comparison with floats is a little weird (you don't have to change anything here, just a note for the future).

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -61,18 +62,25 @@ def run(self,
...
-
-        # Remove this when done
-        raise NotImplementedError
-
+        w_x = self.waypoint.location_x
+        w_y = self.waypoint.location_y
+        r = self.acceptance_radius
+        halted = drone_status.DroneStatus.HALTED
+        move = commands.Command.create_set_relative_destination_command
+        pos = report.position
```
TongguangZhang, line 78: we usually try not to reassign the variable names - this lets other people reading your code not have to memorize the names you've chosen (you don't have to change anything here, just a note).

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -61,18 +62,25 @@ def run(self,
             put_output(command)
         ```
         """
-        # Default command
+        
+        
+       
         command = commands.Command.create_null_command()
-
+
```
TongguangZhang, line 69: No spaces between docstring and code (try not to modify things outside the comment guards). The comment #default command is supposed to go with the code beneath

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -61,18 +82,34 @@ def run(self,
...
+        
+                           
+    
+            
+            
+            
+       
         command = commands.Command.create_null_command()
-
+
```
TongguangZhang, line 93: same as above

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -61,18 +82,34 @@ def run(self,
...
-        raise NotImplementedError
-
+        pos = report.position
+        w_x = self.waypoint.location_x
+        w_y = self.waypoint.location_y
+        r = self.acceptance_radius
+        halted = drone_status.DroneStatus.HALTED
+        move = commands.Command.create_set_relative_destination_command
+        if self.visited == True and report.status == halted:
+            command = commands.Command.create_land_command()
```
TongguangZhang, line 104: What if the drone is halted but not at the landing pad? Make sure you're checking that the drone is halted inside the acceptance radius to the landing pad too.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -36,18 +42,33 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
...
+        i = None
+        for pad in l:
+            dis = self.distance_calculator(startx, starty, pad.location_x, pad.location_y)
+            if mins < 0:
+                mins = dis
+                i = pad
+            elif dis < mins:
+                mins = dis
+                i = pad
+        return i
```
TongguangZhang, line 65: instead of checking if mins has been set each loop, initialize mins to a very large value - either larger than the size of the "map" (unideal), or infinity, with `float("inf")`

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +91,33 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
         # * conf
         # * device
         # * verbose
-        predictions = ...
-
+        confidence_threshold = 0.7
```
TongguangZhang, line 94: set constants like these in the init

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +91,33 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+        predictions = self.__model.predict(image, conf = confidence_threshold, device = DetectLandingPad.__DEVICE, verbose = False)
```
TongguangZhang, line 95: make this multiline, and make sure you're not calling the class with `DetectLandingPad.__DEVICE`, instead use `self.__DEVICE`

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +91,33 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Get the Result object
-        prediction = ...
-
+        prediction =  predictions[0]
         # Plot the annotated image from the Result object
         # Include the confidence value
-        image_annotated = ...
-
+        image_annotated =  prediction.plot()
```
TongguangZhang, line 100: make sure you're including the parameter `conf=True`

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +91,33 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # and convert to a numpy array
-        boxes_cpu = ...
-
+        boxes_cpu = boxes_xyxy.detach().cpu().numpy()
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
+        for box in boxes_cpu:
+            adding_box = bounding_box.BoundingBox.create(box)[1]
+            if adding_box != None:
+                bounding_boxes.append(adding_box)
```
TongguangZhang, line 112: make sure you're not discarding the result flag from the create() call, it tells you if the box creation was successful - if any of the boxes is null, return an empty list instead of skipping that box (use the result flag to determine this)

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -61,18 +62,23 @@ def run(self,
...
+        w_x = self.waypoint.location_x
+        w_y = self.waypoint.location_y
+        r = self.acceptance_radius
+        halted = drone_status.DroneStatus.HALTED
+        move = commands.Command.create_set_relative_destination_command
+        pos = report.position
+        if report.status == halted and self.visited == False:
+            print(self.waypoint)
+            command = move(w_x - pos.location_x, w_y - pos.location_y)
+        if report.status == halted and (w_x - pos.location_x)**2 + (w_y - pos.location_y)**2 <= r**2:
```
TongguangZhang, line 80: make sure to set self.visited = True in the first block and use an elif in the second to prevent unexpected behaviour

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -60,19 +72,26 @@ def run(self,
...
-        # Remove this when done
-        raise NotImplementedError
-
+        pos = report.position
+        w_x = self.waypoint.location_x
+        w_y = self.waypoint.location_y
+        r = self.acceptance_radius
+        halted = drone_status.DroneStatus.HALTED
+        move = commands.Command.create_set_relative_destination_command
+        if self.visited == True and report.status == halted and (self.close.location_x - pos.location_x)**2 + (self.close.location_y - pos.location_y)**2 < r**2:
```
TongguangZhang, line 87: make multiline

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +87,34 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
-        boxes_cpu = ...
-
+        boxes_cpu = boxes_xyxy.detach().cpu().numpy()
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
+        for box in boxes_cpu:
+            added = bounding_box.BoundingBox.create(box)
+            if added[0] == False:
+                return []
+            bounding_boxes.append(added[1])
```
TongguangZhang, line 109: we usually try to split the output with something like
```
result, box = bounding_box.BoundingBox.create(box)
```
so we don't have to use indexing - since it's not always clear what the indices mean

PR #37, opened 2024-02-03
Task1

kevinyvv:


modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,36 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
         # Hint: .shape gets the dimensions of the numpy array
-        # for i in range(0, ...):
-            # Create BoundingBox object and append to list
-            # result, box = ...
+        for i in boxes_numpy:
+            print(i)
+         # Create BoundingBox object and append to list
+            result, box = bounding_box.BoundingBox.create(i)
```
maxlou05, line 113: You have this result variable, but you never used it. Can you find its purpose and implement it? (also just a note: in statistics, if one part of the data point is broken, we discard the entire data point)

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,36 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+            bounding_boxes.append(box)
 
+        return bounding_boxes, image_annotated
         # Remove this when done
-        raise NotImplementedError
+        # raise NotImplementedError
```
maxlou05, line 118: No need to comment this here, you can just delete it. It doesn't explain any code, so git will save the previous versions for you, no need to leave them as comments.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +67,34 @@ def run(self,
...
+                    command = commands.Command.create_land_command()
+            else:
+                self.at_start = False
+                command = commands.Command.create_set_relative_destination_command(required_dist_x, required_dist_y)
+
+        if report.status == drone_status.DroneStatus.MOVING and abs(required_dist_x) < 0.1 and abs(required_dist_y) < 0.1:
+            command = commands.Command.create_halt_command()
         # Remove this when done
-        raise NotImplementedError
+        #raise NotImplementedError
```
maxlou05, line 87: Same, no need to keep

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +67,34 @@ def run(self,
...
+
+        min_dist = 10000 #max possible dist is 60^2 + 60^2, so anything > 7200 works for an initial value
```
maxlou05, line 89: Please do not assume that the max distance is 60, it is not really representative of real life. Use infinity instead.

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -67,11 +67,19 @@ def run(self,
...
+            command = commands.Command.create_land_command()
+        elif report.status == drone_status.DroneStatus.MOVING and abs(required_dist_x) < 0.1 and abs(required_dist_y) < 0.1:
+            command = commands.Command.create_halt_command()
+        else:
+            command = commands.Command.create_set_relative_destination_command(required_dist_x, required_dist_y)
+                
 
         # Remove this when done
-        raise NotImplementedError
+        # raise NotImplementedError
```
maxlou05, line 82: Same, no need to keep

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -67,11 +67,19 @@ def run(self,
         # ============
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
-
+        required_dist_x = self.waypoint.location_x - report.position.location_x
+        required_dist_y = self.waypoint.location_y - report.position.location_y
         # Do something based on the report and the state of this class...
+        if report.status == drone_status.DroneStatus.HALTED and abs(required_dist_x) < 0.1 and abs(required_dist_y) < 0.1:
```
maxlou05, line 73: Good use of the idea of acceptdance radius. However, rather than using a constant here, either set it as a global constant (not ideal) or use the one provided in the function definition.

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -67,11 +67,19 @@ def run(self,
...
+            command = commands.Command.create_land_command()
+        elif report.status == drone_status.DroneStatus.MOVING and abs(required_dist_x) < 0.1 and abs(required_dist_y) < 0.1:
+            command = commands.Command.create_halt_command()
+        else:
+            command = commands.Command.create_set_relative_destination_command(required_dist_x, required_dist_y)
+
```
maxlou05, line 79: A great start, but you just need to add the part to make the drone land after it gets to its destination.

modules/bootcamp/decision_simple_waypoint.py
maxlou05: Additionally, please use euclidean distance rather than taxicab distance.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +67,34 @@ def run(self,
...
 
-        # Do something based on the report and the state of this class...
 
+        required_dist_x = self.waypoint.location_x - report.position.location_x
+        required_dist_y = self.waypoint.location_y - report.position.location_y
+
+
+        # Do something based on the report and the state of this class...
+        if report.status == drone_status.DroneStatus.HALTED:
+            if  abs(required_dist_x) < 0.1 and abs(required_dist_y) < 0.1:
```
maxlou05, line 77: Same comment as from task 3 (use the provided acceptance radius and use euclidean distance)

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +67,34 @@ def run(self,
...
+                if self.at_start == False: # if not at start, then ok to land
+                    command = commands.Command.create_land_command()
```
maxlou05, line 79: A great start, but the bootcamp wants you to fly the drone to land at the nearest landing pad after you get to the waypoint, and not immediately land.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +67,34 @@ def run(self,
...
+            else:
+                self.at_start = False
+                command = commands.Command.create_set_relative_destination_command(required_dist_x, required_dist_y)
+
+        if report.status == drone_status.DroneStatus.MOVING and abs(required_dist_x) < 0.1 and abs(required_dist_y) < 0.1:
+            command = commands.Command.create_halt_command()
```
maxlou05, line 85: I feel like this is going to lag/freeze the drone, because when it's moving towards the waypoint, it's constantly being told to stop, then to move to the waypoint.

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -67,11 +67,19 @@ def run(self,
...
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
-
+        required_dist_x = self.waypoint.location_x - report.position.location_x
+        required_dist_y = self.waypoint.location_y - report.position.location_y
         # Do something based on the report and the state of this class...
+        if report.status == drone_status.DroneStatus.HALTED and abs(required_dist_x) < 0.1 and abs(required_dist_y) < 0.1:
+            command = commands.Command.create_land_command()
+        elif report.status == drone_status.DroneStatus.MOVING and abs(required_dist_x) < 0.1 and abs(required_dist_y) < 0.1:
+            command = commands.Command.create_halt_command()
```
maxlou05, line 76: I feel like this is going to lag/freeze the drone, because when it's moving towards the waypoint, it's constantly being told to stop, then to move to the waypoint.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +67,34 @@ def run(self,
...
         # Remove this when done
-        raise NotImplementedError
+        #raise NotImplementedError
+
+        min_dist = 10000 #max possible dist is 60^2 + 60^2, so anything > 7200 works for an initial value
+        closest_landing_pad = self.waypoint
+
+        for landing_pad_location in landing_pad_locations:
+            #distance can be calculated by sqrt((x1-x2)^2 - (y1-y2)^2), so order is preserved if we do (x1-x2)^2 - (y1-y2)^2 (and saves computation time w/o using sqrt)
```
maxlou05, line 93: Great idea to use distance squared rather than distance! However, you are running the computation twice, it's probably better to save it somewhere.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +67,34 @@ def run(self,
...
+            if (landing_pad_location.location_x - report.position.location_x) ** 2 - (landing_pad_location.location_y - report.position.location_y) ** 2 < min_dist:
+                min_dist = (landing_pad_location.location_x - report.position.location_x) ** 2 - (landing_pad_location.location_y - report.position.location_y) ** 2
+                closest_landing_pad = landing_pad_location
+        self.waypoint = closest_landing_pad
```
maxlou05, line 97: With the current setup, I think the drone will not go to the specified waypoint first, as it is immediately changing the waypoint to the closest landing pad as soon as flight begins. (this function is constantly being called over and over again) Additionally, it is recalculating the closest landing pad constantly, when it only needs to be done once. (you do not need to recalculate the closest landing pad while it's moving toward the closest landing pad)

maxlou05 (commented):
Reviewed

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,36 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
-
-        # Remove this when done
-        raise NotImplementedError
+        for i in boxes_numpy:
+         # Create BoundingBox object and append to list
+            result, box = bounding_box.BoundingBox.create(i)
+            if result:
+                bounding_boxes.append(box)
+            if result == False:
+                bounding_boxes=[]
```
maxlou05, line 116: This is almost there. The code appends bounding boxes onto the picture. by resetting the bounding boxes to empty, it might get populated later again. So instead, you should just return.

maxlou05 (commented):
Reviewed

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +69,39 @@ def run(self,
...
+                if self.closest_pad_found == True: 
+                    # if waypoint has been switched to closest pad, and is within acceptance radius, then it means it has reached pad
+                    command = commands.Command.create_land_command()
+            else:
+                command = commands.Command.create_set_relative_destination_command(required_dist_x, required_dist_y)
+
+
+
+
+        if self.closest_pad_found ==  False and self.at_waypoint == True:
```
maxlou05, line 94: This is good now. However, it's probably clearer if you put it in the condition where you created a null command. It's a little confusing to split one case into two different parts of your code. (also it prevents it from doing an extra if condition check every loop)

maxlou05 (commented):
Reviewed

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,36 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
-
-        # Remove this when done
-        raise NotImplementedError
+        for i in boxes_numpy:
+         # Create BoundingBox object and append to list
+            result, box = bounding_box.BoundingBox.create(i)
+            if result:
+                bounding_boxes.append(box)
+            if result == False:
+                return bounding_boxes, image_annotated
```
maxlou05, line 116: This should be an empty list, as we are discarding the whole thing.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +69,36 @@ def run(self,
...
 
+
+        required_dist_x = self.waypoint.location_x - report.position.location_x
+        required_dist_y = self.waypoint.location_y - report.position.location_y
+
+        distance_away = required_dist_x ** 2 + required_dist_y ** 2
+
         # Do something based on the report and the state of this class...
+        if report.status == drone_status.DroneStatus.HALTED:
+            if  abs(distance_away) < self.acceptance_radius ** 2:
```
maxlou05, line 80: no need for abs() anymore, as squared numbers can't be negative

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -67,11 +67,21 @@ def run(self,
...
+        required_dist_x = self.waypoint.location_x - report.position.location_x
+        required_dist_y = self.waypoint.location_y - report.position.location_y
 
+        #we wont take sqrt of this, and square the acceptance radius instead
+        distance_away = required_dist_x ** 2 + required_dist_y ** 2
+        
         # Do something based on the report and the state of this class...
+        if report.status == drone_status.DroneStatus.HALTED:
+            # if close enough, land at pad
+            if abs(distance_away) < self.acceptance_radius ** 2:
```
maxlou05, line 79: Same here, no need for abs()

maxlou05 (commented):
Reviewed

maxlou05 (approved):
Approved.

PR #38, opened 2024-02-18
Autonomy Bootcamp: Sanjay

sjay05:


modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,31 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
-        # Hint: .shape gets the dimensions of the numpy array
-        # for i in range(0, ...):
-            # Create BoundingBox object and append to list
-            # result, box = ...
+        for box in boxes_cpu:
+            result, bbox = bounding_box.BoundingBox.create(box)
+            if result:
+                bounding_boxes.append(bbox)
```
mgupta27, line 111: If `result=False`, we do not want to take any data from that image.

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -67,11 +68,23 @@ def run(self,
         # ============
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
+        
+        drone_x_coordinate = report.position.location_x
+        drone_y_coordinate = report.position.location_y
+        target_x = self.waypoint.location_x
+        target_y = self.waypoint.location_y
+        distance_squared = (target_x - drone_x_coordinate)**2 + (target_y - drone_y_coordinate)**2
```
mgupta27, line 76: I don't think there is a need to redefine the variables for these values. By creating these variables, we are using more memory than is needed for the program.

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -67,11 +68,23 @@ def run(self,
...
 
-        # Do something based on the report and the state of this class...
-
-        # Remove this when done
-        raise NotImplementedError
+        """
+        Statuses -> Command: 
+         - Halted: Issue new command (either Set Rel. Destination or Land)
+         - Moving: Retain NULL
```
mgupta27, line 81: Good use of docstring!

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -38,6 +38,7 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
         # ============
 
         # Add your own
+        self.epsilon = 1e-1
```
mgupta27, line 41: Use the `acceptance_radius` rather than the epsilon variable.

mgupta27: Same comment for task 4.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -37,12 +37,18 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
...
+        self.epsilon = 1e-1
+        self.closest_landing_pad = None
+        self.reached_waypoint = False
         # Add your own
 
         # ============
         # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
         # ============
 
+    def L2_norm(self, current_loc: location.Location, target_loc: location.Location):
```
mgupta27, line 49: For function names please use lowercase snake_case.

mgupta27: Good use of typing for paramters!

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -67,11 +73,40 @@ def run(self,
         # ============
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
-
-        # Do something based on the report and the state of this class...
-
-        # Remove this when done
-        raise NotImplementedError
+        drone_coordinates = report.position
```
mgupta27, line 76: No need to assign to a new variable.

mgupta27 (commented):
Reviewed.

modules/bootcamp/decision_waypoint_landing_pads.py
sjay05: > For function names please use lowercase snake_case.

I changed it to `l2_norm`.

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,33 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
-        # Hint: .shape gets the dimensions of the numpy array
-        # for i in range(0, ...):
-            # Create BoundingBox object and append to list
-            # result, box = ...
-
-        # Remove this when done
-        raise NotImplementedError
+        for box in boxes_cpu:
+            result, bbox = bounding_box.BoundingBox.create(box)
+            if result == False:
```
TongguangZhang, line 110: A more pythonic expression could be
`if not result:`
you don't need to change anything here, just something to take note of

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,33 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+                return [], image
```
TongguangZhang, line 111: Make sure you're still returning the annotated image here.

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -67,11 +67,19 @@ def run(self,
...
+        """
+        Statuses -> Command: 
+         - Halted: Issue new command (either Set Rel. Destination or Land)
+         - Moving: Retain NULL
+        """
+        if report.status == drone_status.DroneStatus.HALTED:
+            if distance_squared < (self.acceptance_radius**2):
+                command = commands.Command.create_land_command()
+            else:
+                command = commands.Command.create_set_relative_destination_command(self.waypoint.location_x - report.position.location_x, self.waypoint.location_y - report.position.location_x)
```
TongguangZhang, line 82: Make this line multiline since it exceeds 100 characters.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +73,38 @@ def run(self,
...
 
-        # Do something based on the report and the state of this class...
-
-        # Remove this when done
-        raise NotImplementedError
+        if self.reached_waypoint:
+            # Reached waypoint, find closest landing location
+
+            if self.closest_landing_pad == None:
+                assert len(landing_pad_locations) > 0
```
TongguangZhang, line 80: We don't want to halt the program if we get an empty list of landing_pad_locations, we want to return the default null command instead

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +73,38 @@ def run(self,
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
-        # Do something based on the report and the state of this class...
-
-        # Remove this when done
-        raise NotImplementedError
+        if self.reached_waypoint:
```
TongguangZhang, line 76: We only want to update the command when the drone is stopped - consider altering the structure of your if statements. For example - if we haven't reached the waypoint, go to the waypoint, if we're at the waypoint, find the closest landing pad and go to it, if we're at the landing pad, land

TongguangZhang (commented):
Great job Sanjay, your logic for waypoint landing pads seems a little complicated, but everything else looks great!

modules/bootcamp/detect_landing_pad.py
sjay05: Thanks, fixed.

modules/bootcamp/decision_simple_waypoint.py
sjay05: Thanks. I set it to a multi-line function call.

modules/bootcamp/decision_waypoint_landing_pads.py
sjay05: Ah okay. I removed the assert statement so that it would return null by default.

modules/bootcamp/decision_waypoint_landing_pads.py
sjay05: I reorganized the if structure to make it more natural as you mentioned and the command should only be updated when the drone is halted.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +73,46 @@ def run(self,
...
+                distance_to_pad = self.l2_norm(report.position, self.closest_landing_pad)
+
+                if report.status == drone_status.DroneStatus.HALTED:
+                    if distance_to_pad < self.acceptance_radius**2:
+                        command = commands.Command.create_land_command()
+                    else:
+                        command = commands.Command.create_set_relative_destination_command(
+                            self.closest_landing_pad.location_x - report.position.location_x, 
+                            self.closest_landing_pad.location_y - report.position.location_y
+                        )
```
TongguangZhang, line 115: Your logic here works - but there is some unideal behaviour. We only need to do something if the drone is halted - so that should always be the first check, not if it's reached the waypoint or if there's a landing pad, since we won't do anything with that if the drone isn't halted.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +73,46 @@ def run(self,
...
+                        closest_landing_pad = lpad
+
+                self.closest_landing_pad = closest_landing_pad
+
+            elif report.status == drone_status.DroneStatus.HALTED:
+                # Closest landing pad has been found. Land.
+
+                distance_to_pad = self.l2_norm(report.position, self.closest_landing_pad)
+
+                if report.status == drone_status.DroneStatus.HALTED:
```
TongguangZhang, line 108: This if statement is repeated, we already know that the drone is halted in this block

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +73,46 @@ def run(self,
...
+                # If closest landing pad was not calculated, then find it and go to it.
+                closest_pad_distance = float('inf')
+                closest_landing_pad = None
+                for lpad in landing_pad_locations:
+                    calculated_distance = self.l2_norm(report.position, lpad)
+                    if calculated_distance < closest_pad_distance:
+                        closest_pad_distance = calculated_distance
+                        closest_landing_pad = lpad
+
+                self.closest_landing_pad = closest_landing_pad
```
TongguangZhang, line 101: We should look for the landing pad as soon as we get to the waypoint, not in the next iteration of the loop

TongguangZhang (commented):
LGTM!

PR #39, opened 2024-02-23
Completed Autonomy bootcamp.

aritrakar:
# Overview

Completed the Autonomy bootcamp for W24 semester (based on F32 bootcamp).

# References

https://uwarg-docs.atlassian.net/wiki/spaces/BOOT/pages/2299756545/Autonomy+Bootcamp+-+F23

### DIFF
--- a/modules/bootcamp/detect_landing_pad.py
+++ b/modules/bootcamp/detect_landing_pad.py
@@ -113,7 +113,6 @@ class DetectLandingPad:
         # Hint: .shape gets the dimensions of the numpy array
         for i in range(boxes_cpu.shape[0]):
             # Create BoundingBox object and append to list
-            print(boxes_cpu[i])
             result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
             if result:
                 bounding_boxes.append(box)
@@ -124,3 +123,5 @@ class DetectLandingPad:
         # ============
         # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
         # ============
+    
+        return command
--- a/modules/bootcamp/detect_landing_pad.py
+++ b/modules/bootcamp/detect_landing_pad.py
@@ -86,32 +86,40 @@ class DetectLandingPad:
         # * conf
         # * device
         # * verbose
-        predictions = ...
+        predictions = self.__model.predict(
+            source=image,
+            conf=0.7,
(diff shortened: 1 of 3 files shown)


modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,33 +86,42 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
         # Hint: .shape gets the dimensions of the numpy array
-        # for i in range(0, ...):
+        for i in range(boxes_cpu.shape[0]):
             # Create BoundingBox object and append to list
-            # result, box = ...
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+            if result:
+                bounding_boxes.append(box)
```
mgupta27, line 118: If `result = False`, then we don't want to collect any data from that image source. If there is an error with the data, we want to remove the entire data point which is the image source.

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,33 +86,42 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
-        # Remove this when done
-        raise NotImplementedError
+        # Return the list of bounding boxes and the annotated image
+        return bounding_boxes, image_annotated
 
         # ============
         # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
         # ============
+    
+        return command
```
mgupta27, line 127: Can you remove these 2 lines.

mgupta27 (commented):
Reviewed.

modules/bootcamp/detect_landing_pad.py
aritrakar: Sure. I left it there because bootcampers were asked not to touch it and it wouldn't be executed anyway.

modules/bootcamp/detect_landing_pad.py
aritrakar: So should I return an empty list and the original image if `result = False`?

mgupta27: Yes.

mgupta27 (commented):
LGTM!

PR #41, opened 2024-03-06
bootcamp

nrudba:


modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -37,7 +37,7 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
-        # Add your own
+        self.eps = 0.01
```
maxlou05, line 40: Great idea. But just use the given `acceptance_radius` rather than creating your own.

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,33 +86,30 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
-        # Hint: .shape gets the dimensions of the numpy array
-        # for i in range(0, ...):
-            # Create BoundingBox object and append to list
-            # result, box = ...
-
-        # Remove this when done
-        raise NotImplementedError
+        for i in range(boxes_cpu.shape[0]):
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+            if result: bounding_boxes.append(box)
```
maxlou05, line 110: Great job! However, usually we would discard the entire dataset if part of it is wrong (in this case, we would discard the entire image, ie. return [] instead of some of the valid bounding boxes). (You do not need to change anything here, just something to think about)

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -67,11 +69,33 @@ def run(self,
...
-        # Remove this when done
-        raise NotImplementedError
+        def sqrdist(p1: location.Location, p2: location.Location):
+            dx = p2.location_x - p1.location_x
+            dy = p2.location_y - p1.location_y
+            return dx * dx + dy * dy
+
+        if self.reached_waypoint:
+            if not self.landing_pad:
+                self.landing_pad = min(landing_pad_locations, key = lambda pt: sqrdist(pt, self.waypoint))
```
maxlou05, line 79: Great use of the min and lambda!

maxlou05 (commented):
Reviewed.

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,10 +68,20 @@ def run(self,
...
-        # Do something based on the report and the state of this class...
-
-        # Remove this when done
-        raise NotImplementedError
+        def sqrdist(p1: location.Location, p2: location.Location):
+            dx = p2.location_x - p1.location_x
+            dy = p2.location_y - p1.location_y
+            return dx * dx + dy * dy
+
+        if report.status != drone_status.DroneStatus.HALTED: return command
```
maxlou05, line 76: There is no need to return an empty command when the drone is not halted. You can just remove this line, as constantly sending a new move command is fine as well.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -67,11 +67,26 @@ def run(self,
...
-        # Do something based on the report and the state of this class...
-
-        # Remove this when done
-        raise NotImplementedError
+        def sqrdist(p1: location.Location, p2: location.Location):
+            dx = p2.location_x - p1.location_x
+            dy = p2.location_y - p1.location_y
+            return dx * dx + dy * dy
+
+        if report.status != drone_status.DroneStatus.HALTED: return command
```
maxlou05, line 75: Same here, you can just keep sending move commands, no need to send an empty command.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -67,11 +67,26 @@ def run(self,
...
+
+        if self.landing_pad:
```
maxlou05, line 77: You need to add back the check for when the drone is close to the landing pad. This will instantly land as soon as we decide which landing pad to land at.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -67,11 +67,26 @@ def run(self,
...
+            command = commands.Command.create_land_command()
+        elif sqrdist(report.position, self.waypoint) <= self.acceptance_radius * self.acceptance_radius:
+            self.landing_pad = min(landing_pad_locations, key=lambda pt: sqrdist(pt, self.waypoint))
+            command = commands.Command.create_set_relative_destination_command(
```
maxlou05, line 81: You may want to add a check in case `self.landing_pad = None`, as it would give an error here.

maxlou05 (commented):
Reviewed.

maxlou05 (commented):
LGTM

PR #43, opened 2024-04-02
Alvin Autonomy Bootcamp

AJM432:


modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -37,7 +37,12 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
-        # Add your own
+        # Validate data
+        if abs(self.waypoint.location_x) > 60 or abs(self.waypoint.location_y) > 60:
+            print("Invalid waypoint, must be in flight boundary")
+    
+    def euclidean_distance_squared(self, x0: float, y0: float, x1: float, y1: float) -> float:
+        return (x1-x0)**2 + (y1-y0)**2
```
maxlou05, line 45: Great use of type hinting and distance squared! (No need to change anything here)

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,34 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
         # * conf
         # * device
         # * verbose
-        predictions = ...
+        predictions = self.__model.predict(source=image, conf=0.7, device=self.__DEVICE, verbose=True)
```
maxlou05, line 89: You could turn off the logs if you want (ie set verbose to False). (You do not need to change anything here)

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,34 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
         # Hint: .shape gets the dimensions of the numpy array
-        # for i in range(0, ...):
+        for i in range(boxes_cpu.shape[0]):
             # Create BoundingBox object and append to list
-            # result, box = ...
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+            if not result:
+                return [], image_annotated
```
maxlou05, line 113: Great job recognizing that we should discard the entire image when part of it has issues! (You do not need to change anything here)

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -36,8 +36,34 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
...
+
+    def euclidean_distance_squared(self, x0: float, y0: float, x1: float, y1: float) -> float:
+        return (x1-x0)**2 + (y1-y0)**2
+    
+
+    def get_closest_landing_pad(self,
+        report: drone_report.DroneReport,
+        landing_pad_locations: "list[location.Location]") -> location.Location:
+        closest_distance_to_landing_squared = float('inf')
+        closest_landing_location = landing_pad_locations[0]
```
maxlou05, line 55: While the bootcamp specifies at least 1 landing pad, what happens if there were no landing pads? (You do not need to change anything here, just something to think about).

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -36,8 +36,34 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
...
+
+        for location in landing_pad_locations:
```
maxlou05, line 57: Also, there is no need to initialize `closest_landing_pad` since you are going to compare it with itself here first anyways.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +94,32 @@ def run(self,
...
+        if report.status == drone_status.DroneStatus.HALTED:
+            # Only compute distance if needed within halt branch
+            distance_to_destination_squared = self.euclidean_distance_squared(
+                report.position.location_x, report.position.location_y, 
+                self.destination.location_x, self.destination.location_y)
+
+            # Reached the destination
+            if distance_to_destination_squared < self.acceptance_radius**2:
+                # Accounts for the case when waypoint is on landing pad and/or when reaching landing pad
+                if self.reached_waypoint or self.waypoint in landing_pad_locations:
```
maxlou05, line 107: With your current control flow, you do not actually need to check if the waypoint is on landing pad since it must be on the landing pad when `self.reached_waypoint` is True (you set it to true only after the destination becomes the landing pad). (You do not need to change anything here, just something to think about)

maxlou05 (commented):
LGTM!

PR #44, opened 2024-04-04
Bootcamp Complete - Pranav Bommireddipalli

officialpranav:


modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -67,11 +72,36 @@ def run(self,
...
+        else:
+            #If drone has reached landing pad, land
+            if self.waypoint_is_landing_pad:
+                command = commands.Command.create_land_command()
+                return command
+            
+            #Find closest landing pad
+            self.waypoint_is_landing_pad = True
+            
+            closest_landing_pad_location: location.Location = None
```
TongguangZhang, line 91: good use of type annotation here! (you don't have to change anything here, I just really liked it)

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -67,11 +72,36 @@ def run(self,
...
-
-        # Remove this when done
-        raise NotImplementedError
+        
+        #If drone is already moving to waypoint do nothing
+        if report.status == drone_status.DroneStatus.MOVING:
+            return command
+               
+        if self.__distance_squared(self.waypoint.location_x, self.waypoint.location_y, report.position.location_x, report.position.location_y) > self.acceptance_radius**2 :
+            command = commands.Command.create_set_relative_destination_command(self.waypoint.location_x - report.position.location_x, self.waypoint.location_y - report.position.location_y)
```
TongguangZhang, line 81: Your logic is good for this specific case, but consider if the drone halts for a second time within the radius of the first waypoint - this would make the drone land on the waypoint instead of the landing pad. Let's try not modifying self.waypoint and instead use a state to determine whether or not the waypoint has been visited yet.

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,33 +86,35 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
 
         # Get the xyxy boxes list from the Boxes object in the Result object
-        boxes_xyxy = ...
+        boxes_xyxy = prediction.boxes.xyxy
 
         # Detach the xyxy boxes to make a copy,
         # move the copy into CPU space,
         # and convert to a numpy array
-        boxes_cpu = ...
+        boxes_cpu = boxes_xyxy.cpu().numpy()
```
TongguangZhang, line 103: make sure you call detach() on boxes_xyxy to make a copy rather than using the boxes themselves, the reason for this is that the boxes are tensors with gradients on for the purposes of training the model. Since we're only using the model output, we want gradients off.

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,33 +86,35 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Hint: .shape gets the dimensions of the numpy array
-        # for i in range(0, ...):
-            # Create BoundingBox object and append to list
-            # result, box = ...
-
-        # Remove this when done
-        raise NotImplementedError
-
+        for i in range(0, boxes_cpu.shape[0]):
+            box = bounding_box.BoundingBox.create(boxes_cpu[i])
```
TongguangZhang, line 110: let's use result, box to make it more clear what the variable is, since index 0 of the result is not a bounding box, but a success flag

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,33 +86,35 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+            if box[0] is False:
+                continue
```
TongguangZhang, line 112: In statistics, we actually want to delete a data point instead of returning a partial data point if it is corrupted. This means that we want to return nothing (an empty list) if any box creation fails.

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,33 +86,35 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+            
+            bounding_boxes.append(box[1])
+
+        return((bounding_boxes, image_annotated))
```
TongguangZhang, line 116: let's try to avoid duplicate brackets here, `return bounding boxes, image_annotated` will work here

TongguangZhang (commented):
Reviewed! There are a couple edge cases which you might not be covered in your logic, but it is largely good

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,33 +86,35 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
-
+        predictions = self.__model.predict(image, conf=0.7, device=self.__DEVICE,verbose=False)
         # Get the Result object
-        prediction = ...
+        prediction = predictions[0]
 
         # Plot the annotated image from the Result object
         # Include the confidence value
-        image_annotated = ...
+        image_annotated = prediction.plot()
```
TongguangZhang, line 95: we want to set a confidence threshold for the annotated image, with conf = True

TongguangZhang (commented):
reviewed!

TongguangZhang (commented):
LGTM!

PR #45, opened 2024-04-04
autonomy bootcamp Ernest Wang

ErnestWang31:


modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,36 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+        boxes_cpu = boxes_xyxy.detach().cpu().numpy()
 
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
         # Hint: .shape gets the dimensions of the numpy array
         # for i in range(0, ...):
-            # Create BoundingBox object and append to list
-            # result, box = ...
+        # Create BoundingBox object and append to list
+        # result, box = ...
```
TongguangZhang, line 111: Make sure you delete commented out code - if we need it we can go back in the commit history for it.

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,36 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+
+        for i in range (0, boxes_cpu.shape[0]):
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+            bounding_boxes.append(box)
```
TongguangZhang, line 115: What happens if the result is False, should we still append the bounding box?

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -69,12 +70,28 @@ def run(self,
         # ============
 
         # Do something based on the report and the state of this class...
-
+        if report.status == drone_status.DroneStatus.HALTED:
+            if self.relative_distance(report) > self.acceptance_radius:
+                command = commands.Command.create_set_relative_destination_command(self.waypoint.location_x
+                                                                                   - report.position.location_x,
+                                                                                   self.waypoint.location_y
+                                                                                   - report.position.location_y)
```
TongguangZhang, line 78: Take a look at the python style guide for multiline code https://uwarg-docs.atlassian.net/wiki/spaces/CV/pages/2253226033/Python+Style+Convention#Multiline.
Or, hot tip, use the black formatter, documentation here https://uwarg-docs.atlassian.net/wiki/spaces/CV/pages/2473361462/Linters+and+formatters#Black

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -69,12 +70,28 @@ def run(self,
...
+            else:
+                command = commands.Command.create_land_command()
+        elif report.status == drone_status.DroneStatus.MOVING:
+            if self.relative_distance(report) < self.acceptance_radius:
+                command = commands.Command.create_halt_command()
```
TongguangZhang, line 83: I like the extra check here (you don't need to change anything here)

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -69,12 +70,28 @@ def run(self,
...
+        # raise NotImplementedError
 
         # ============
         # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
         # ============
 
         return command
+
+    def relative_distance(self, report: drone_report.DroneReport) -> float:
+        r_d = ((self.waypoint.location_x - report.position.location_x) ** 2
```
TongguangZhang, line 94: We try to avoid variable names like r_d, since it's hard to tell what that means, maybe "distance" would e a better name

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -69,12 +70,28 @@ def run(self,
...
+               + (self.waypoint.location_y - report.position.location_y) ** 2) ** 0.5
```
TongguangZhang, line 95: Note that squaring the acceptance radius instead of square rooting the relative distance will give us the same results when we compare, but squaring is a cheaper operation that square rooting (you don't have to change anything here, just something to think about)

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -20,10 +20,13 @@
 
 # All logic around the run() method
 # pylint: disable-next=too-few-public-methods
+
+
```
TongguangZhang, line 24: Make sure you're not changing anything outside the comment guards (this extra whitespace is also against the style convention)

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -20,10 +20,13 @@
...
 class DecisionWaypointLandingPads(base_decision.BaseDecision):
     """
     Travel to the designed waypoint and then land at the nearest landing pad.
     """
+
```
TongguangZhang, line 29: Same as above

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -70,11 +74,61 @@ def run(self,
 
         # Do something based on the report and the state of this class...
 
+        if self.waypoint_status:  # if at waypoint
```
TongguangZhang, line 77: Do we need to calculate the closest landing pad on every run if the drone isn't halted? Consider when the command needs to change

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -70,11 +74,61 @@ def run(self,
...
+            # land at nearest pad
+            position = report.position
```
TongguangZhang, line 79: This line is in both the if and the else (line 97) - consider putting it outside the if condition if you'll always need it

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -70,11 +74,61 @@ def run(self,
...
+            landing_pad = self.find_nearest_landing_pad(position, landing_pad_locations)
+
+            if report.status == drone_status.DroneStatus.HALTED:
+                # if halted -> go to pad
+                if self.get_relative_distance(position, landing_pad) > self.acceptance_radius:
+                    report.destination = landing_pad
```
TongguangZhang, line 85: Since you never use report.destination, we don't need to assign it here

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -70,11 +74,61 @@ def run(self,
...
         # ============
         # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
         # ============
 
         return command
+
+    def find_nearest_landing_pad(self, position: location.Location,
+                                 landing_pad_locations: "list[location.Location]") -> location.Location:
+
+        nearest_landing_pad = landing_pad_locations[0]
```
TongguangZhang, line 123: what if the list is empty? It might be better to initialize this to None

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -70,11 +74,61 @@ def run(self,
...
+        for landingPad in landing_pad_locations:
+            distance = self.get_relative_distance(position, landingPad)
+            if distance < nearest_distance:
+                nearest_landing_pad = landingPad
+                nearest_distance = distance
+        return nearest_landing_pad
+
+    def get_relative_distance(self, position: location.Location, destination: location.Location) -> float:
+        return ((position.location_x - destination.location_x) ** 2 +
+                (position.location_y - destination.location_y) ** 2)
```
TongguangZhang, line 134: if you're not using the square root here, remember to square the acceptance radius

TongguangZhang (commented):
Reviewed! Logic is mostly good - just a little reorganizing in `decision_waypoint_landing_pads.py` and some style changes

modules/bootcamp/decision_example.py
TongguangZhang, line 1: This file shouldn't be changed

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -43,9 +45,11 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
...
         # ============
 
-    def run(self,
-            report: drone_report.DroneReport,
-            landing_pad_locations: "list[location.Location]") -> commands.Command:
+    def run(
+        self,
+        report: drone_report.DroneReport,
+        landing_pad_locations: "list[location.Location]",
+    ) -> commands.Command:
```
TongguangZhang, line 52: I would say to not change things outside the comment guards, but good on you for looking at the new style convention (you don't need to change anything here).

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +92,37 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
-        # Remove this when done
-        raise NotImplementedError
+        for i in range(0, boxes_cpu.shape[0]):
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+            if result:
+                bounding_boxes.append(box)
+            else:
+                print("Failed to create bounding box")
+                bounding_boxes = []
+                break
```
TongguangZhang, line 123: Let's try to do early exiting here
```
if not result:
    exit
append box
```

TongguangZhang (commented):
Reviewed! Almost there

TongguangZhang (commented):
LGTM!

PR #47, opened 2024-04-04
Completed bootcamp tasks

chloe778:


modules/bootcamp/detect_landing_pad.py
```diff
@@ -27,7 +27,7 @@ class DetectLandingPad:
     # Chooses the GPU if it exists, otherwise runs on the CPU
     # If you have a CUDA capable GPU but want to force it to
     # run on the CPU instead, replace the right side with "cpu"
-    __DEVICE = 0 if torch.cuda.is_available() else "cpu"
+    __DEVICE = "cpu"
```
TongguangZhang, line 30: This code is supposed to be device agnostic - we want to use CUDA if we can, and CPU only if it's unavailable (also make sure you're not modifying anything outside of the comment guards)

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,36 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
 
         # Get the xyxy boxes list from the Boxes object in the Result object
-        boxes_xyxy = ...
+        boxes_xyxy = prediction.boxes.xyxy
 
         # Detach the xyxy boxes to make a copy,
         # move the copy into CPU space,
         # and convert to a numpy array
-        boxes_cpu = ...
+        boxes_cpu = boxes_xyxy.cpu().numpy()
```
TongguangZhang, line 104: Make sure you call detach() on boxes_xyxy, since we want a copy of the tensor without tracking the gradients

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,36 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
 
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
+        for i in range(boxes_cpu.shape[0]):
+            bounds = boxes_cpu[i, :4]
```
TongguangZhang, line 109: Try not to index with magic numbers - we can't be sure if the size of the bounding box is actually 4.

I'm guessing you're using `:4` since it returns a copy of the box instead of the actual box? in that case `:` is safer, but also if you call detach() you don't need to make another copy since it's already a copy.

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,36 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+            success, box = bounding_box.BoundingBox.create(bounds)
+            if success:
+                bounding_boxes.append(box)
```
TongguangZhang, line 112: Good job checking the success flag! However, instead of skipping invalid bounding boxes, we actually want to return an empty list if any of the bounding boxes are invalid

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,36 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Hint: .shape gets the dimensions of the numpy array
         # for i in range(0, ...):
             # Create BoundingBox object and append to list
             # result, box = ...
 
-        # Remove this when done
-        raise NotImplementedError
+        return(bounding_boxes, image_annotated)
```
TongguangZhang, line 118: try to avoid unnecessary brackets - `return bounding_boxes, image_annotated` will work here

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -67,12 +67,19 @@ def run(self,
         # ============
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
-
+        current_pos = report.position
+        distance_sqr = (self.waypoint.location_x - current_pos.location_x) ** 2 + \
+            (self.waypoint.location_y - current_pos.location_y) ** 2
+        acceptance_radius_sqr = self.acceptance_radius ** 2
```
TongguangZhang, line 73: We don't need to calculate this on every run, since we only use these values if the drone is halted.

TongguangZhang, line 73: since the acceptance_radius won't change, the accpetance_radius_sqr won't either, declare it in the __init__() as self.acceptance_radius_squared to avoid calculating it each time

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -67,12 +67,19 @@ def run(self,
...
+        if report.status == drone_status.DroneStatus.HALTED:
+            if distance_sqr <= acceptance_radius_sqr:
+                command = commands.Command.create_land_command()
+            else:
+                x = self.waypoint.location_x - current_pos.location_x
+                y = self.waypoint.location_y - current_pos.location_y
```
TongguangZhang, line 79: since these are the relative and not absolute x and y positions, it might be better to name them as such to avoid confusion

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -37,11 +37,25 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
...
-        # Add your own
+        self.at_waypoint = False
+        self.closest_landing_pad = None
 
         # ============
         # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
         # ============
+    def distance_sqr (self, p1: location.Location, p2: location.Location):
+        return ((p1.location_x - p2.location_x) ** 2 + \
+        (p1.location_y - p2.location_y) ** 2)
```
TongguangZhang, line 48: Since this fits on one line (less than 100 characters) let's keep it one line

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -37,11 +37,25 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
...
+    
+    def find_closest_pad (self, p1: location.Location, landing_pad_locations: "list[location.Location]"):
+        min_dist = float("inf")
+        best_location = landing_pad_locations[0]
```
TongguangZhang, line 52: What if the list is empty? It might be better to initialize it to None here

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -64,14 +78,26 @@ def run(self,
         # Default command
         command = commands.Command.create_null_command()
 
+
```
TongguangZhang, line 81: Try to avoid modifying lines outside the comment guards

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -64,14 +78,26 @@ def run(self,
...
         # ============
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
+        if self.at_waypoint and not self.closest_landing_pad:
+            self.closest_landing_pad = self.find_closest_pad(report.position, landing_pad_locations)
```
TongguangZhang, line 86: Search for the closest landing pad below inside the else at least 93 to avoid the duplicated if statement

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -64,14 +78,26 @@ def run(self,
...
+        if report.status == drone_status.DroneStatus.HALTED:
+            if not self.at_waypoint:
+                if self.distance_sqr(self.waypoint, report.position) <= self.acceptance_radius * self.acceptance_radius:
+                    self.at_waypoint = True
+                else:
+                    command = commands.Command.create_set_relative_destination_command((self.waypoint.location_x - report.position.location_x), (self.waypoint.location_y- report.position.location_y))
```
TongguangZhang, line 92: Make sure you remove duplicate brackets
Missing space on `(self.waypoint.location_y- report.position.location_y)`

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -64,14 +78,26 @@ def run(self,
...
+            if not self.at_waypoint:
+                if self.distance_sqr(self.waypoint, report.position) <= self.acceptance_radius * self.acceptance_radius:
+                    self.at_waypoint = True
+                else:
+                    command = commands.Command.create_set_relative_destination_command((self.waypoint.location_x - report.position.location_x), (self.waypoint.location_y- report.position.location_y))     
+            else:
+                if self.distance_sqr(self.closest_landing_pad, report.position) <= self.acceptance_radius * self.acceptance_radius:
+                    command = commands.Command.create_land_command()
+                else: 
+                    command = commands.Command.create_set_relative_destination_command((self.closest_landing_pad.location_x - report.position.location_x), (self.closest_landing_pad.location_y - report.position.location_y))
```
TongguangZhang, line 97: Remove duplicate brackets here

TongguangZhang (commented):
Reviewed

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,36 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         bounding_boxes = []
+        for i in range(boxes_cpu.shape[0]):
+            bounds = boxes_cpu[i, :4]
+            success, box = bounding_box.BoundingBox.create(bounds)
+            if success:
+                bounding_boxes.append(box)
         # Hint: .shape gets the dimensions of the numpy array
         # for i in range(0, ...):
             # Create BoundingBox object and append to list
             # result, box = ...
```
TongguangZhang, line 116: Make sure you delete commented out code - we can go back in the git history if we need it later

TongguangZhang (commented):
LGTM!

PR #48, opened 2024-04-06
Momin Bootcamp

Momin-C:


modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,44 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
-        image_annotated = ...
-
-        # Get the xyxy boxes list from the Boxes object in the Result object
-        boxes_xyxy = ...
+        path = pathlib.Path("models/")
+        createResult, instance = self.create(path)
+        if not createResult:
+            raise ValueError
+        
+        model = instance.__model
```
TongguangZhang, line 95: we want to use self._model here - this is for encapsulation purposes

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,44 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
 
-        # Detach the xyxy boxes to make a copy,
-        # move the copy into CPU space,
-        # and convert to a numpy array
-        boxes_cpu = ...
+        image_path = "modules/bootcamp/"
```
TongguangZhang, line 97: we don't need to save the images - just return the annotated image

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,44 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
-        # Hint: .shape gets the dimensions of the numpy array
-        # for i in range(0, ...):
-            # Create BoundingBox object and append to list
-            # result, box = ...
-
-        # Remove this when done
-        raise NotImplementedError
+        # Get the Result object
+        results = model.predict(image, verbose = False, conf = 0.7)
+        for result in results:
```
TongguangZhang, line 101: we don't need to iterate of the results - we're only predicting on 1 image so there will only be 1 result

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,44 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+            
+            # Plot the annotated image from the Result object
+            # Include the confidence value
+            sum = np.sum(image)
```
TongguangZhang, line 105: don't redefine sum here (it's a python keyword)

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,44 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+            image_file = f"result{sum}.jpg"
+            result.save(filename=image_path + image_file)
```
TongguangZhang, line 107: we don't want to save the image here

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,44 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+            result_array = result.plot()
```
TongguangZhang, line 108: we want to use a confidence threshold here, with the parameter conf = True

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,44 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+
+            # Get the xyxy boxes list from the Boxes object in the Result object
+            boxes_xyxy = result.boxes.xyxy
+
+            # Detach the xyxy boxes to make a copy,
+            # move the copy into CPU space,
+            # and convert to a numpy array
+            boxes_cpu = boxes_xyxy.cpu().numpy()
```
TongguangZhang, line 116: make sure you call boxes_xyxy.detach() - this makes a copy of the results that is detached from the computational graph. We want a copy because the original has gradients on (which is used for the back propagation in training the model) but since we're just predicting, we don't want the gradients

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,44 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+
+            # Loop over the boxes list and create a list of bounding boxes
+            bounding_boxes = []
+            
+            for box in boxes_cpu:
+                boxResult, boundingBox = bounding_box.BoundingBox.create(box)
+                if boxResult:
+                    bounding_boxes.append(boundingBox)
```
TongguangZhang, line 124: we actually don't just want to skip a failed box creation, we want skip the whole image (by returning an empty list)

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -67,11 +68,33 @@ def run(self,
...
+        #Restricting to flight boundary
+        if to_travel_x > 60:
+            to_travel_x = 60
+        elif to_travel_x < -60:
+            to_travel_x = -60
+        
+        if to_travel_y > 60:
+            to_travel_y = 60
+        elif to_travel_y < -60:
+            to_travel_y = -60
```
TongguangZhang, line 87: how'd you get these values? I don't think the images are 120x120 but I could be wrong
we also don't want to set it to the max values if the values exceed the bounds - the waypoint isn't at the bounds

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -67,11 +68,33 @@ def run(self,
         # ============
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
+        
+        position_x, position_y = report.position.location_x, report.position.location_y
+        destination_x, destination_y = self.waypoint.location_x, self.waypoint.location_y
```
TongguangZhang, line 73: let's try to not reassign the variable names, use report.position.location_x, etc. throughout so it's easier to read (and uses less memory)

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -67,11 +68,33 @@ def run(self,
...
+        
+        if to_travel_y > 60:
+            to_travel_y = 60
+        elif to_travel_y < -60:
+            to_travel_y = -60
 
         # Do something based on the report and the state of this class...
+        dist_squared = to_travel_x**2 + to_travel_y**2
+        if dist_squared <= self.radius_squared:
+            command = commands.Command.create_land_command()
```
TongguangZhang, line 92: we don't want the drone to land if it's not halted (it could move unpredictably)

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -67,11 +68,33 @@ def run(self,
...
+        elif report.status == drone_status.DroneStatus.HALTED:
+            command = commands.Command.create_set_relative_destination_command(to_travel_x, to_travel_y)
 
         # Remove this when done
-        raise NotImplementedError
+        # raise NotImplementedError
```
TongguangZhang, line 97: let's remove commented out or "dead" code - we can go back in the git history for it if we ever need it again (just to be clear, we do want to keep annotations explaining what the code does, just not actual code itself)

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -38,6 +38,9 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
         # ============
 
         # Add your own
+        self.radius_squared = self.acceptance_radius**2
+        self.reachedWayPoint = False
+        self.toLandAt = waypoint # Initializing to waypoint
```
TongguangZhang, line 43: we try to use snake_case, not camelCase in our code (it's also good to keep it consistent within your own code)

TongguangZhang, line 43: we don't want to land at the waypoint - initialize to None instead

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,13 +71,59 @@ def run(self,
...
-        raise NotImplementedError
+        # raise NotImplementedError
 
         # ============
         # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
         # ============
 
         return command
+    
+    def distance(self, start: location.Location, end: location.Location) -> float:
```
TongguangZhang, line 123: technically this is the distance squared - let's rename it to be more descriptive

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,13 +71,59 @@ def run(self,
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
+        position_x, position_y = report.position.location_x, report.position.location_y
```
TongguangZhang, line 74: similar to decision_simple_waypoint.py

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,13 +71,59 @@ def run(self,
...
+        #Restricting to flight boundary
+        if to_travel_x > 60:
+            to_travel_x = 60
+        elif to_travel_x < -60:
+            to_travel_x = -60
+        
+        if to_travel_y > 60:
+            to_travel_y = 60
+        elif to_travel_y < -60:
+            to_travel_y = -60
```
TongguangZhang, line 93: similar to decision_simple_waypoint.py

TongguangZhang (commented):
reviewed

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,38 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
         # * conf
         # * device
         # * verbose
-        predictions = ...
 
+        path = pathlib.Path("models/")
+        createResult, _ = self.create(path)
+        if not createResult:
+            raise ValueError
+
```
TongguangZhang, line 94: we don't need to create an instance of this object inside the run function - let's avoid raising errors unless we know the thing calling it can catch the error

TongguangZhang (commented):
reviewed

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,33 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # * device
         # * verbose
-        predictions = ...
-
+        
         # Get the Result object
-        prediction = ...
-
+        result = self.__model.predict(image, verbose = False, conf = 0.7)[0]
+
```
TongguangZhang, line 92: let's get the predictions object first - we're not guaranteed that the index 0 exists, also, include the device argument in the call

TongguangZhang (commented):
reviewed

TongguangZhang (commented):
LGTM!

PR #50, opened 2024-04-15
Bootcamp - Tao Shen

ShiftyTS:


modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,25 +86,37 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
 
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
+
+        # Get the dimensions of the numpy array
+        num_boxes, _ = boxes_cpu.shape
+        
+        for i in range(num_boxes):
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+            bounding_boxes.append(box)
```
TongguangZhang, line 116: what if the bounding box creation fails (i.e. result is False), do we still want to append the box?

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -67,14 +67,33 @@ def run(self,
         # ============
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
-
-        # Do something based on the report and the state of this class...
-
-        # Remove this when done
-        raise NotImplementedError
-
+        status = report.status
```
TongguangZhang, line 70: let's not reassign variables - use report.status throughout so it's easier to keep track of (and uses less memory)

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -67,14 +67,33 @@ def run(self,
...
+            if self.relative_distance(relative_x_dist, relative_y_dist) < self.acceptance_radius:
+                command = commands.Command.create_land_command()
+            else:
+                command = commands.Command.create_set_relative_destination_command(
+                    relative_x_dist,
+                    relative_y_dist,
+                )
+        elif status == drone_status.DroneStatus.MOVING:
+            if self.relative_distance(relative_x_dist, relative_y_dist) < self.acceptance_radius:
+                command = commands.Command.create_halt_command()
```
TongguangZhang, line 85: I like this extra check (you don't have to change anything here but combining both if statements with an and would reduce the redundancy)

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -67,14 +67,70 @@ def run(self,
...
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
-
-        # Do something based on the report and the state of this class...
-
-        # Remove this when done
-        raise NotImplementedError
-
+        status = report.status
+        position = report.position
```
TongguangZhang, line 71: same as above, let's not reassign variables - use report.status and report.position throughout so it's easier to keep track of (and uses less memory)

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -67,14 +67,70 @@ def run(self,
...
+
+        if self.visited_waypoint: # Travel to nearest landing pad
+            if status == drone_status.DroneStatus.HALTED: # If halted, need to first calculate the nearest landing pad
+                self.nearest_landing_pad = self.get_nearest_landing_pad(position, landing_pad_locations)
```
TongguangZhang, line 75: avoid calculating this twice (when you reach the landing pad you run this again)

TongguangZhang (commented):
Reviewed

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -67,14 +67,70 @@ def run(self,
...
+        nearest_landing_pad = None
+        nearest_landing_pad_distance = float("inf")
+        for landing_pad in landing_pad_locations:
+            current_distance = self.get_relative_location_distance(position, landing_pad)
+            if current_distance < nearest_landing_pad_distance:
+                nearest_landing_pad = landing_pad
+                nearest_landing_pad_distance = current_distance
+        return nearest_landing_pad
+    
+    def get_relative_location_distance(
```
TongguangZhang, line 131: this is actually the squared distance - let's rename this function to be more descriptive

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -67,14 +67,70 @@ def run(self,
...
+            self, 
+            locOne: location.Location, 
+            locTwo: location.Location,
```
TongguangZhang, line 134: we try to use snake_case, not camelCase

PR #51, opened 2024-04-28
Autonomy Bootcamp - Ivy Ye

47vy:


modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -69,9 +76,19 @@ def run(self,
         # ============
 
         # Do something based on the report and the state of this class...
+        
+        if report.status == drone_status.DroneStatus.HALTED:
+            distance = self.distance_to_waypoint(report.position)
+            if distance <= (self.acceptance_radius ** 2):
```
TongguangZhang, line 82: try to avoid duplicate brackets here

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -38,11 +38,18 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
         # ============
 
         # Add your own
-
+        
         # ============
         # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
         # ============
 
+    def distance_to_waypoint(self, current_position: location.Location) -> float:
```
TongguangZhang, line 46: since this is the squared distance, let's rename it to be more descriptive

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,9 +91,33 @@ def run(self,
         # ============
 
         # Do something based on the report and the state of this class...
+        if report.status == drone_status.DroneStatus.HALTED:
+            if not self.reached:
+                distance_to_waypoint_squared = self.distance_to_waypoint(report.position, self.waypoint)
+                
+                if distance_to_waypoint_squared < self.acceptance_radius**2:
+                    self.reached = True
+                    return commands.Command.create_land_command()
```
TongguangZhang, line 100: we don't want to land on the waypoint - we want to search for landing pads after reaching the waypoint, either do not return a command here or move the code below that searches for landing pads to here instead

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,33 +86,39 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Get the xyxy boxes list from the Boxes object in the Result object
-        boxes_xyxy = ...
+        boxes_xyxy = prediction.boxes.xyxy
 
-        # Detach the xyxy boxes to make a copy,
+        # Detach the xyxy boxes to make a copy,s
         # move the copy into CPU space,
         # and convert to a numpy array
-        boxes_cpu = ...
+        boxes_cpu = boxes_xyxy.cpu().numpy()
```
TongguangZhang, line 104: we want to call detach() on this as well, which removes it from the computational graph
tensors attached to the graph have gradients on, which is used for back propagation during model training, since we're just using the model for predictions, we want to have the gradients off

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,33 +86,39 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Hint: .shape gets the dimensions of the numpy array
         # for i in range(0, ...):
+        # print(boxes_cpu.shape[0])
+        for i in range(0, boxes_cpu.shape[0]):
             # Create BoundingBox object and append to list
             # result, box = ...
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+            print(box)
+            if result:
+                bounding_boxes.append(box)
```
TongguangZhang, line 117: instead of skipping a failed box creation, we actually want to return nothing for the whole image

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -69,9 +76,19 @@ def run(self,
...
+                command = commands.Command.create_land_command()
+            else:
+                command = commands.Command.create_set_relative_destination_command(
+                    self.waypoint.location_x - report.position.location_x, 
+                    self.waypoint.location_y - report.position.location_y)
+
 
         # Remove this when done
-        raise NotImplementedError
+        # raise NotImplementedError
```
TongguangZhang, line 91: we want to avoid having commented out code since it won't be run, if we need this code we can go back in the git history for it (we do want to keep annotative comments though)

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,9 +91,33 @@ def run(self,
...
+                    best_pad = self.best_landing_pad(report.position, landing_pad_locations)
+                    self.landing_pad_location = best_pad
+
+                    return commands.Command.create_set_relative_destination_command(
+                        self.landing_pad_location.location_x - report.position.location_x,
+                        self.landing_pad_location.location_y - report.position.location_y)
 
         # Remove this when done
-        raise NotImplementedError
+        # raise NotImplementedError
```
TongguangZhang, line 120: same as above

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,33 +86,39 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
 
         # Remove this when done
-        raise NotImplementedError
-
+        # raise NotImplementedError
+        return bounding_boxes, image_annotated
```
TongguangZhang, line 121: we want to avoid having commented out code since it won't be run, if we need this code we can go back in the git history for it (we do want to keep annotative comments though)

TongguangZhang (commented):
Great job! Overall logic looks good, just some minor fixes

TongguangZhang (commented):
LGTM!

PR #52, opened 2024-05-01
Bootcamp Completed

aniketsrinivasan:


modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +70,46 @@ def run(self,
...
 
-        # Do something based on the report and the state of this class...
-
-        # Remove this when done
-        raise NotImplementedError
+        # Finds the closest landing pad to the drone's current location:
+        def distance_to_pad_sqr(pad_loc: location.Location, drone_loc: location.Location) -> float:
+            distance_sqr = ((drone_loc.location_x - pad_loc.location_x) ** 2 +
+                            (drone_loc.location_y - pad_loc.location_y) ** 2)
+            return distance_sqr
```
TongguangZhang, line 77: let's avoid defining functions within other functions - it'll declare this function every time run() is called, but we only want to declare it once when the object is initialized

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -37,7 +37,9 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
-        # Add your own
+        # Keeping track of what stage the drone is at:
+        self.at_pad = True
```
TongguangZhang, line 41: the variable name here is a little confusing - in your code it sounds like
if at the landing pad and not at the waypoint, go to the waypoint
if not at the landing pad and at the waypoint, find and go to the landing pad
if not at the landing pad and not at the waypoint, land

some of it seems reversed to me

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +70,46 @@ def run(self,
...
+                command = commands.Command.create_set_relative_destination_command(
+                    best_pad.location_x - report.position.location_x,
+                    best_pad.location_y - report.position.location_y
+                )
+                self.at_waypoint = False
+            # Task 3: Land.
+            #   not at initial pad
+            #   not at final waypoint
+            elif not self.at_pad and not self.at_waypoint:
+                command = commands.Command.create_land_command()
```
TongguangZhang, line 112: make sure you check the acceptance radius to make sure you are at the waypoint or at the landing pad

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +70,46 @@ def run(self,
...
+                    self.waypoint.location_x - report.position.location_x,
+                    self.waypoint.location_y - report.position.location_y)
+                self.at_pad = False
+                self.at_waypoint = True
+            # Task 2: Move from waypoint to the closest landing pad.
+            #   not at initial pad
+            #   at final waypoint
+            elif not self.at_pad and self.at_waypoint:
+                # Finding the closest landing pad location:
+                best_pad = landing_pad_locations[0]
```
TongguangZhang, line 95: what if the list is empty? (we should initialize this variable to None)

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +70,46 @@ def run(self,
...
+                best_dist_sqr = distance_to_pad_sqr(best_pad, self.waypoint)
```
TongguangZhang, line 96: we're not guaranteed to be exactly at the waypoint, especially if the drone stops in the middle - we should use the position passed into the function instead

TongguangZhang, line 96: we shouldn't initialize this to the first pad distance, since you're calculating it again below, initialize to infinity instead

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,34 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
         # * conf
         # * device
         # * verbose
-        predictions = ...
+        predictions = self.__model.predict(image, conf=0.7, \
+                                           device=(0 if torch.cuda.is_available() else "cpu"), \
+                                           verbose=True)
```
TongguangZhang, line 91: you don't need the slashes here (refer to the python style guide for multiline code) https://uwarg-docs.atlassian.net/wiki/spaces/CV/pages/2253226033/Python+Style+Convention

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,34 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
         # * conf
         # * device
         # * verbose
-        predictions = ...
+        predictions = self.__model.predict(image, conf=0.7, \
+                                           device=(0 if torch.cuda.is_available() else "cpu"), \
```
TongguangZhang, line 90: use the constant here for device

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,34 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
-        # for i in range(0, ...):
-            # Create BoundingBox object and append to list
-            # result, box = ...
 
-        # Remove this when done
-        raise NotImplementedError
+        for i in range(0, boxes_cpu.shape[0]):
+            box = boxes_cpu[i]
+            result, bbox = bounding_box.BoundingBox.create(box)
+            bounding_boxes.append(bbox)
```
TongguangZhang, line 114: what if the result is false, do we still want to append the bounding box?

TongguangZhang (commented):
Reviewed

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,10 +74,22 @@ def run(self,
...
+            print(self.counter)
+            print(self.command_index)
+            print("Halted at: " + str(report.position))
+
+            command = self.commands[self.command_index]
+            self.command_index += 1
+        elif report.status == drone_status.DroneStatus.HALTED and not self.has_sent_landing_command:
+            command = commands.Command.create_land_command()
+
+            self.has_sent_landing_command = True
```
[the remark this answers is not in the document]
aniketsrinivasan: right, thanks for pointing this out. resolved.

modules/bootcamp/decision_waypoint_landing_pads.py
aniketsrinivasan: that's a good point—i've fixed this.

modules/bootcamp/decision_waypoint_landing_pads.py
aniketsrinivasan: ah, i see your point. i've played around with the variable names now, should make more sense

modules/bootcamp/decision_waypoint_landing_pads.py
aniketsrinivasan: right, resolved.

modules/bootcamp/decision_waypoint_landing_pads.py
aniketsrinivasan: oh, because we don't know if landing_pad_locations is empty. i've resolved this. ty!

modules/bootcamp/decision_waypoint_landing_pads.py
aniketsrinivasan: resolved.

aniketsrinivasan: i don't think it was being calculated again below (iteration was through `landing_pad_locations[1:]` but this makes sense to fix considering we now initialize `best_pad` to `None`. fixed.

modules/bootcamp/detect_landing_pad.py
aniketsrinivasan: oops, forgot to remove these! thanks

modules/bootcamp/detect_landing_pad.py
aniketsrinivasan: resolved.

modules/bootcamp/detect_landing_pad.py
aniketsrinivasan: ah, i see. if the result is false, then we don't want to append, and can return None instead since we have no boxes in that case. resolved.

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,38 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
 
-        # Remove this when done
-        raise NotImplementedError
+        for i in range(0, boxes_cpu.shape[0]):
+            box = boxes_cpu[i]
+            result, bbox = bounding_box.BoundingBox.create(box)
+            if result:
+                bounding_boxes.append(bbox)
+            else:
+                return None
```
AaronWang04, line 118: returning None here and returning bounding_boxes, image_annotated is an error.

someone trying to unpack the result of this function will expect 2 results (ex. x, y = func() ).  if func returns 1 result it will cause an error

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,10 +74,25 @@ def run(self,
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
-        # Do something based on the report and the state of this class...
-
-        # Remove this when done
-        raise NotImplementedError
+        # Modified "decision_example" to work for just one command.
```
AaronWang04, line 77: i would also add some comments about how if status is moving, the default command is returned, just for sake of clarity for other readers. otherwise lgtm

AaronWang04 (commented):
just some final comments, almost there!

modules/bootcamp/detect_landing_pad.py
aniketsrinivasan: hmm, so then we should return `[], None` since `bounding_boxes` would be empty, in this case. or we can just return `bounding_boxes, None` since we've already initialized it. fixed.

modules/bootcamp/decision_simple_waypoint.py
aniketsrinivasan: i'll do this, thank you. resolved.

AaronWang04 (approved):
looks good to me!

PR #55, opened 2024-05-06
Andrew Shum - completed bootcamp tasks

ashum68:


modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +73,30 @@ def run(self,
...
         # ============
 
-        # Do something based on the report and the state of this class...
+        relative_x = self.waypoint.location_x - report.position.location_x
+        relative_y = self.waypoint.location_y - report.position.location_y
+
+        if report.status == drone_status.DroneStatus.HALTED:
+            if not self.at_waypoint:
+                command = commands.Command.create_set_relative_destination_command(relative_x, relative_y)
+                self.at_waypoint = True
```
TongguangZhang, line 82: make sure to check that the drone has reached the waypoint before setting `self.at_waypoint` to be true

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +73,30 @@ def run(self,
...
+            else:
+                min_pad_dist = 10000
+                closest = 0
```
TongguangZhang, line 85: better to use object instead of index (list could be empty when accessing)

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +73,30 @@ def run(self,
...
+                for i in range(len(landing_pad_locations)):
+                    pad_dist = self.dist_to_pad(landing_pad_locations[i])
+                    if pad_dist < min_pad_dist:
+                        min_pad_dist = pad_dist
+                        closest = i
+                if min_pad_dist < 10000:
```
TongguangZhang, line 91: check if object is not None instead

TongguangZhang (commented):
Reviewed

TongguangZhang (commented):
LGTM!

PR #57, opened 2024-05-14
Junzhang Pull Request

JunzhL:


modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,35 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+        predictions = self.__model.predict(source=image, conf=0.7, device=self.__DEVICE, verbose=False)
 
         # Get the Result object
-        prediction = ...
+        prediction = predictions[0]
 
         # Plot the annotated image from the Result object
         # Include the confidence value
-        image_annotated = ...
+        image_annotated = prediction.plot()
```
maxlou05, line 96: Please include the confidence value, as indicated. (hint, it's conf)

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,35 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Hint: .shape gets the dimensions of the numpy array
         # for i in range(0, ...):
             # Create BoundingBox object and append to list
             # result, box = ...
-
-        # Remove this when done
-        raise NotImplementedError
+        for i in range(boxes_cpu.shape[0]):
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+            if result:
```
maxlou05, line 114: In statistics, an invalid portion of the data point involves removing the entirety of the data point (where the data point here is an image, as each image is independent). (You do not need to change anything, just something to think about)

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -38,11 +38,23 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
         # ============
 
         # Add your own
+        self.command_index = 0
+        self.commands = [
+            commands.Command.create_set_relative_destination_command(waypoint.location_x, waypoint.location_y),
```
maxlou05, line 43: Please note that this command is a relative destination command, and not absolute coordinates. You should take into account your own position.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,12 +81,38 @@ def run(self,
...
-        # Remove this when done
-        raise NotImplementedError
+        if report.status == drone_status.DroneStatus.HALTED and self.command_index < len(self.commands):
+            # Print some information for debugging
+            print(self.counter)
+            print(self.command_index)
+            print("Halted at: " + str(report.position))
+
+            command = self.commands[self.command_index]
+            self.has_arrived_at_waypoint = True
```
maxlou05, line 91: This code assumes that you are guaranteed to make it to the destination, which is specified in this bootcamp. What would happen if the drone halts somewhere in between? (You do not need to change anything, just something to think about)

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,12 +81,38 @@ def run(self,
...
+        elif report.status == drone_status.DroneStatus.HALTED and self.has_sent_landing_command:
+
+            command = commands.Command.create_land_command()
+
+        elif report.status == drone_status.DroneStatus.HALTED and self.has_arrived_at_waypoint:
+            nearest_landing_pad = None
+            nearest_distance = None
+
+            for landing_pad in landing_pad_locations:
+                distance = ((report.position.location_x - landing_pad.location_x) ** 2 + (report.position.location_y - landing_pad.location_y) ** 2) ** 0.5
```
maxlou05, line 103: The square root function is expensive, is there any way to determine the closest landing pad without the square root?

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -39,6 +39,15 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
 
         # Add your own
 
+        self.command_index = 0
+        self.commands = [
+            commands.Command.create_set_relative_destination_command(waypoint.location_x, waypoint.location_y),
```
maxlou05, line 44: Please note that this command is a relative destination command, and not absolute coordinates. You should take into account your own position.

maxlou05 (commented):
Reviewed

modules/bootcamp/detect_landing_pad.py
maxlou05: If you looked at the documentation, it should be `plot(conf=True)`

maxlou05 (commented):
Reviewed

maxlou05 (commented):
LGTM!

PR #60, opened 2024-05-20
Alex Le's Pull Request

AlexLe68:


modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,38 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
 
         # Get the xyxy boxes list from the Boxes object in the Result object
-        boxes_xyxy = ...
+        boxes_xyxy = prediction.boxes.xyxy
 
         # Detach the xyxy boxes to make a copy,
         # move the copy into CPU space,
         # and convert to a numpy array
-        boxes_cpu = ...
+        boxes_cpu = boxes_xyxy.cpu().numpy()
```
TongguangZhang, line 104: make sure you call detach() to remove it from the computational graph
basically, this turns of the gradients, gradients are default on since they're used in for back propagation in model training, but since we're only inferencing here, we don't want to keep track of the gradients

TongguangZhang (commented):
reviewed

TongguangZhang (commented):
LGTM!

PR #66, opened 2024-05-28
Autonomy Bootcamp ready for review (xuan pan)

Xuan0ling:


modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,41 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+        predictions = self.__model.predict(source=image, conf=0.5, device=self.__DEVICE, verbose=True)
 
         # Get the Result object
-        prediction = ...
+        prediction = predictions[0]
 
         # Plot the annotated image from the Result object
         # Include the confidence value
-        image_annotated = ...
+        image_annotated = prediction.plot()
```
maxlou05, line 96: Please include the confidence level in the image annotations (use the conf kwarg)

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,41 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # move the copy into CPU space,
         # and convert to a numpy array
-        boxes_cpu = ...
+        boxes_cpu = boxes_xyxy.detach().cpu().numpy()
 
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
+        for i in range (boxes_cpu.shape[0]):
+            result, bounding_Box_obj = bounding_box.BoundingBox.create(boxes_cpu[i])
+            if (result):
```
maxlou05, line 111: Great job recognizing that we discard the entire image (datapoint) if even one part of it is defective. Although, you do not need the brackets here. (You do not need to change anything)

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -39,6 +39,15 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
 
         # Add your own
 
+        self.command_index = 0
+        self.commands = [
+            commands.Command.create_set_relative_destination_command(waypoint.location_x, waypoint.location_y)
```
maxlou05, line 44: Please note that this is a relative destination, your own position matters.

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -69,12 +78,22 @@ def run(self,
...
         # Do something based on the report and the state of this class...
+        if report.status == drone_status.DroneStatus.HALTED:
+            # Print some information for debugging
+            print(self.counter)
+            print(self.command_index)
+            print("Halted at: " + str(report.position))
+
+            command = self.commands[self.command_index]
+            self.command_index += 1
+        elif report.status == drone_status.DroneStatus.HALTED and not self.has_sent_landing_command:
```
maxlou05, line 89: I'm pretty sure that this will never run, since it's an else if. It will only check this condition if the first if fails, and if the first if fails that means the drone is not halted, thus this code never runs.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -39,6 +39,18 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
 
         # Add your own
 
+        self.command_index = 0
+        self.commands = [
+            commands.Command.create_set_relative_destination_command(waypoint.location_x, waypoint.location_y)
```
maxlou05, line 44: Again, please note that this is a relative position, and your own position matters

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,9 +81,37 @@ def run(self,
...
+            print("Halted at: " + str(report.position))
+
+            
+            command = self.commands[self.command_index]
+
+            self.go_to_landing_pad = True
+        elif report.status == drone_status.DroneStatus.HALTED and self.go_to_landing_pad and not self.has_sent_landing_command:
+            nearest_landing_pad = landing_pad_locations[0]
+            min_distance = ((nearest_landing_pad.location_x - report.position.location_x)**2 + (nearest_landing_pad.location_y - report.position.location_y)**2)**0.5
+            print("123456789")
```
maxlou05, line 97: What is the purpose of this?

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,9 +81,37 @@ def run(self,
...
+            print(self.command_index)
+            print("Halted at: " + str(report.position))
+
+            
+            command = self.commands[self.command_index]
+
+            self.go_to_landing_pad = True
+        elif report.status == drone_status.DroneStatus.HALTED and self.go_to_landing_pad and not self.has_sent_landing_command:
+            nearest_landing_pad = landing_pad_locations[0]
+            min_distance = ((nearest_landing_pad.location_x - report.position.location_x)**2 + (nearest_landing_pad.location_y - report.position.location_y)**2)**0.5
```
maxlou05, line 96: Is there a way to avoid computing the square root? Think about what is your purpose for computing the distance.

modules/bootcamp/decision_waypoint_landing_pads.py
Xuan0ling: I was debugging and forgot to delete it.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,9 +81,37 @@ def run(self,
...
+            print("123456789")
+            for landing_pad in landing_pad_locations:
```
maxlou05, line 98: Since `nearest_landing_pad` was already initialized to the first element of the list, it's not necessary to compare it with itself. Think about how you can optimize this.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,9 +81,37 @@ def run(self,
...
+                distance = ((landing_pad.location_x - report.position.location_x)**2 + (landing_pad.location_y - report.position.location_y)**2)**0.5
```
maxlou05, line 99: Again, think about how you can optimize this by not computing the square root.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,9 +81,37 @@ def run(self,
...
+            print(self.counter)
+            print(self.command_index)
+            print("Halted at: " + str(report.position))
+
+            
+            command = self.commands[self.command_index]
+
+            self.go_to_landing_pad = True
+        elif report.status == drone_status.DroneStatus.HALTED and self.go_to_landing_pad and not self.has_sent_landing_command:
+            nearest_landing_pad = landing_pad_locations[0]
```
maxlou05, line 95: This is fine, although in practice you have to think about what if `landing_pad_locations` is an empty list. (You do not need to change anything, just something to think about)

maxlou05 (commented):
Reviewed

modules/bootcamp/decision_simple_waypoint.py
maxlou05: Now both of the if and else if statements are exactly the same, which will again result in the else if not running. I think you just forgot to change the `not`

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,9 +81,36 @@ def run(self,
...
+            print(self.counter)
+            print(self.command_index)
+            print("Halted at: " + str(report.position))
+
+            
+            command = self.commands[self.command_index]
+
+            self.go_to_landing_pad = True
+        elif report.status == drone_status.DroneStatus.HALTED and self.go_to_landing_pad and not self.has_sent_landing_command:
+            nearest_landing_pad = commands.Command.create_set_relative_destination_command(self.waypoint.location_x - report.position.location_x, self.waypoint.location_y - report.position.location_y)
```
maxlou05, line 95: I think you put this on the wrong line? It should probably be on line 91

maxlou05 (commented):
Reviewed

maxlou05 (commented):
Approved!

PR #67, opened 2024-05-30
first

ZetongYao:


modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,45 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+        )
 
         # Get the Result object
-        prediction = ...
+        prediction = predictions[0]
 
         # Plot the annotated image from the Result object
         # Include the confidence value
-        image_annotated = ...
+        image_annotated = prediction.plot()
```
maxlou05, line 101: Please include the confidence value in the image (use the conf kwarg)

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,45 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # for i in range(0, ...):
             # Create BoundingBox object and append to list
             # result, box = ...
-
-        # Remove this when done
-        raise NotImplementedError
+        
+        for i in range(boxes_cpu.shape[0]):
+            results, bbox = bounding_box.BoundingBox.create(boxes_cpu[i])
+            if not results:
```
maxlou05, line 120: Normally in statistics, we would discard the entire data point (the whole image and all the bounding boxes) in case one part of it (one bounding box) fails. (You do not need to change anything, just something to think about)

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,12 +72,52 @@ def run(self,
...
-        raise NotImplementedError
+        if report.status == drone_status.DroneStatus.HALTED:
+            if not self.traveled_to_waypoint:
+                # Fly to the waypoint
+                command = commands.Command.create_set_relative_destination_command(
+                    self.waypoint.location_x - report.position.location_x,
+                    self.waypoint.location_y - report.position.location_y
+                )
+                self.traveled_to_waypoint = True
+            elif not self.closest_landing_pad:
```
maxlou05, line 83: Using this as a statement to check if you have found the closest landing pad or not can be a little dangerous, in the sense that if there were no landing pads in the list (ie it was empty), you wouldn't know that you had already checked that it was empty. (You do not need to change anything, just something to think about)

maxlou05 (commented):
Reviewed

PR #69, opened 2024-06-08
Raiyan Samin Bootcamp

OhRai:


modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,39 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
 
         # Get the xyxy boxes list from the Boxes object in the Result object
-        boxes_xyxy = ...
+        boxes_xyxy = prediction.boxes.xyxy
 
         # Detach the xyxy boxes to make a copy,
         # move the copy into CPU space,
         # and convert to a numpy array
-        boxes_cpu = ...
+        boxes_cpu = boxes_xyxy.cpu().numpy()
```
maxlou05, line 104: Also detach the xyxy boxes with 'detach()'

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,39 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
         # Hint: .shape gets the dimensions of the numpy array
-        # for i in range(0, ...):
+        for i in range(0, len(boxes_cpu)):
             # Create BoundingBox object and append to list
-            # result, box = ...
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+
+            if result:
```
maxlou05, line 114: Usually, if there's an issue with one part of the data (one bounding box in this image), we discard the whole data point. (You do not need to change anything here, just something to think about)

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -70,11 +70,22 @@ def run(self,
 
         # Do something based on the report and the state of this class...
 
+        # If halted, move to the waypoint
+        if report.position != self.waypoint and report.status == drone_status.DroneStatus.HALTED:
```
maxlou05, line 74: Try not to use the condition that your location is exactly equal to the waypoint, as this is a very difficult thing to achieve. Normally, there are errors associated with these values, and it would be extremely difficult for them to be exactly the same. Instead, try thinking about when it's "close enough".

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -70,11 +70,22 @@ def run(self,
...
+            command = commands.Command.create_set_relative_destination_command(self.waypoint.location_x, self.waypoint.location_y)
```
maxlou05, line 75: This is a relative destination command, while the waypoint is an absolute position. Thus, your own position matters and will change where you end up going.

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -70,11 +70,22 @@ def run(self,
...
+            # print(command)
+
+        # If at the waypoint, land
+        elif report.position == self.waypoint:
```
maxlou05, line 79: See previous comment about using exact position comparisons.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -38,11 +38,19 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
...
 
         # ============
         # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
         # ============
 
+    def _get_distance(self, location1: location.Location, location2: location.Location) -> float:
+        """
+        Get the distance between two locations.
+        """
+        return math.sqrt(pow((location1.location_x - location2.location_x), 2) + pow((location1.location_y - location2.location_y), 2))
```
maxlou05, line 52: Think about how you can not use the square root function, which takes a lot of computing time. Think about what is your end purpose for calculating distances.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -70,8 +79,36 @@ def run(self,
 
         # Do something based on the report and the state of this class...
 
+        # print(report)
+
+        # If halted, move to the waypoint
+        if (report.position != self.waypoint) and (report.status == drone_status.DroneStatus.HALTED) and (not self.reached_waypoint):
```
maxlou05, line 85: See comment about position

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -70,8 +79,36 @@ def run(self,
...
+            command = commands.Command.create_set_relative_destination_command(self.waypoint.location_x, self.waypoint.location_y)
```
maxlou05, line 86: See comment about relative distance

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -70,8 +79,36 @@ def run(self,
...
+
+        # If at the waypoint, find the nearest landing pad
+        if (report.position == self.waypoint) and (not self.reached_waypoint):
```
maxlou05, line 89: See comment about position comparison

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -70,8 +79,36 @@ def run(self,
...
+            command = commands.Command.create_halt_command()
+            self.reached_waypoint = True
+
+            # Find the nearest landing pad
+            for location in landing_pad_locations:
+                distance = self._get_distance(report.position, location)
```
maxlou05, line 95: Although this works, you are calculating distance twice every loop. You can actually just save the minimum distance you've seen thus far and compare with that, just like how you would find a minimum value in an array in a linear search (You do not have to change anything here, just something to think about)

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -70,8 +79,36 @@ def run(self,
...
+                    self.nearest_landing_pad = location
+
+            print("Nearest Landing Pad: ", self.nearest_landing_pad)
+
+            # Move the nearest landing pad
+            relative_x = self.nearest_landing_pad.location_x - report.position.location_x
+            relative_y = self.nearest_landing_pad.location_y - report.position.location_y
+            command = commands.Command.create_set_relative_destination_command(relative_x=relative_x, relative_y=relative_y)
+
+        elif (report.position == self.nearest_landing_pad) and (self.reached_waypoint):
```
maxlou05, line 106: see comment about position comparison

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -70,8 +79,36 @@ def run(self,
...
+            command = commands.Command.create_halt_command()
+            command = commands.Command.create_land_command()
```
maxlou05, line 108: see comment about duplicate commands

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -70,8 +79,36 @@ def run(self,
...
+
         # Remove this when done
-        raise NotImplementedError
+        # raise NotImplementedError
```
maxlou05, line 111: No need to keep old code as a comment, you can just delete it. This is what Git is for.

maxlou05 (commented):
Reviewed

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -70,11 +70,22 @@ def run(self,
...
+            command = commands.Command.create_halt_command()
+            command = commands.Command.create_land_command()
```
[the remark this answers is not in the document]
OhRai: Ah I see. I thought I would have to halt it the drone once it reached the waypoint.

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -43,6 +43,12 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
         # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
         # ============
 
+    def _get_distance(self, location1: location.Location, location2: location.Location) -> float:
+        """
+        Get the distance between two locations.
+        """
+        return ((location1.location_x - location2.location_x) ** 2 + (location1.location_y - location2.location_y) ** 2) ** 0.5
```
maxlou05, line 50: Just compute distance squared and compare using squared distances

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -70,8 +78,37 @@ def run(self,
...
+            if self._get_distance(report.position, self.waypoint) < self.acceptance_radius:
+                if not self.reached_waypoint:
+                    self.reached_waypoint = True
+                    # Find the nearest landing pad
+                    for location in landing_pad_locations:
+                        distance = self._get_distance(report.position, location)
+                        if (self.nearest_landing_pad is None) or (distance < self._get_distance(report.position, self.nearest_landing_pad)):
+                            self.nearest_landing_pad = location
+
+                else:
```
maxlou05, line 93: There seems to be an issue with the logic here: the drone should land at the landing pad location, but this landing command can only be run if the drone is at the waypoint location (since it's inside the waypoint's location checking if statement)

maxlou05 (commented):
Reviewed

modules/bootcamp/decision_waypoint_landing_pads.py
maxlou05: I don't think that you need this else clause here, since you should also send the command to go to the nearest landing pad as soon as you reach the waypoint. This would still work since the function is just going to run again in a very short period of time, but it is a little more confusing to read.

maxlou05 (commented):
Reviewed

maxlou05 (commented):
Approved!

PR #70, opened 2024-06-19
Autonomy bootcamp pull request

Mhorse2471:


modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +85,37 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+        predictions = self.__model.predict(source=image, conf=0.7, device=self.__DEVICE, verbose=True)
 
         # Get the Result object
-        prediction = ...
+        prediction = predictions[0]
 
         # Plot the annotated image from the Result object
         # Include the confidence value
-        image_annotated = ...
+        image_annotated = prediction.plot()
```
maxlou05, line 95: Please include the confidence value with kwarg `conf`

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +85,37 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # and convert to a numpy array
-        boxes_cpu = ...
+        boxes_cpu = boxes_xyxy.detach().cpu().numpy()
 
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
+        for i in range(boxes_cpu.shape[0]):
+            print(boxes_cpu[i])
+            box = boxes_cpu[i]
+            bounding_boxes.append(bounding_box.BoundingBox.create(bounds=np.array([box[0], box[1], box[2], box[3]]))[1])
```
maxlou05, line 110: Be aware that we sometimes creating a bounding box will fail, in which case we will reject the image. Please use the

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -64,14 +63,28 @@ def run(self,
...
 
-        # Do something based on the report and the state of this class...
+        #Send destination command
+        command = commands.Command.create_set_relative_destination_command(self.waypoint.location_x, self.waypoint.location_y)
+
+        if (self.on_route == False):
+            command = commands.Command.create_set_relative_destination_command(self.waypoint.location_x, self.waypoint.location_y)
+            self.on_route = True
+
+        elif (report.status.value == 0):
```
maxlou05, line 79: Please use the enum instead of the integer value (ie `report.status == drone_status.DroneStatus.HALTED`)

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -12,7 +12,7 @@
 from .. import drone_status
 from .. import location
 from ..private.decision import base_decision
-
+import math
```
maxlou05, line 15: Please do not import other libraries. (You don't actually need the square root function to complete this bootcamp)

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -64,14 +63,28 @@ def run(self,
...
 
+
+
         # ============
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
-        # Do something based on the report and the state of this class...
+        #Send destination command
+        command = commands.Command.create_set_relative_destination_command(self.waypoint.location_x, self.waypoint.location_y)
```
maxlou05, line 73: Please note that this is a relative destination command, while the waypoint location is absolute. Your own position will affect where you end up going.

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -64,14 +63,28 @@ def run(self,
...
+
+        if (self.on_route == False):
+            command = commands.Command.create_set_relative_destination_command(self.waypoint.location_x, self.waypoint.location_y)
```
maxlou05, line 76: Same as above: relative distance required

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -64,17 +63,44 @@ def run(self,
...
+
         # ============
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
         # Do something based on the report and the state of this class...
 
+        #Step 1: Travel to waypoint
+        if (self.step == 0):
+            command = commands.Command.create_set_relative_destination_command(self.waypoint.location_x, self.waypoint.location_y)
```
maxlou05, line 77: relative distance required

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -64,17 +63,44 @@ def run(self,
...
+            self.step = 1
+        elif (self.step == 1 and report.status.value == 1):
+            self.closest_pad = landing_pad_locations[0]
+            # for i in range (landing_pad_locations):
+            if (getDist(landing_pad_locations[0].location_x, landing_pad_locations[0].location_y, report.position.location_x, report.position.location_y) < getDist(self.closest_pad.location_x, self.closest_pad.location_y, report.position.location_x, report.position.location_y)):
```
maxlou05, line 82: Not sure what's going on here, it seems that you are comparing the same thing with itself? I think you meant to find the closest landing pad here with the for loop...

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -64,17 +63,44 @@ def run(self,
...
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
         # Do something based on the report and the state of this class...
 
+        #Step 1: Travel to waypoint
+        if (self.step == 0):
+            command = commands.Command.create_set_relative_destination_command(self.waypoint.location_x, self.waypoint.location_y)
+            self.step = 1
+        elif (self.step == 1 and report.status.value == 1):
```
maxlou05, line 79: Again, please use the enum constants instead of the integer value

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -64,17 +63,44 @@ def run(self,
...
+            self.closest_pad = landing_pad_locations[0]
+            # for i in range (landing_pad_locations):
+            if (getDist(landing_pad_locations[0].location_x, landing_pad_locations[0].location_y, report.position.location_x, report.position.location_y) < getDist(self.closest_pad.location_x, self.closest_pad.location_y, report.position.location_x, report.position.location_y)):
+                self.closest_pad = landing_pad_locations[0]
+            command = commands.Command.create_set_relative_destination_command(self.closest_pad.location_x - report.position.location_x, self.closest_pad.location_y - report.position.location_y)
+            print("Moving")
+            self.step = 2
+        elif (self.step == 2 and report.status.value == 1):
```
maxlou05, line 87: Please use the enum variable

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -64,17 +63,44 @@ def run(self,
...
+        
         # ============
         # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
         # ============
-
         return command
+
+
+def getDist(x1, y1, x2, y2):
+    return math.sqrt((x1 - x2)**2 + (y1 - y2)**2)
```
maxlou05, line 106: Note that you do not need the square root function

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -64,14 +63,28 @@ def run(self,
...
+            self.on_route = True
+
+        elif (report.status.value == 0):
+            command = commands.Command.create_null_command()
+        else:
```
maxlou05, line 81: How do you know here that the drone has finished moving to the waypoint? Although in this simulation the drone will always make it to the destination, it might not in real life. Please include a check that you are sufficiently near the waypoint. (hint: `self.acceptance_radius` exists as a parameter for this reason)

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +85,37 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+        return bounding_boxes, image_annotated
         # Hint: .shape gets the dimensions of the numpy array
-        # for i in range(0, ...):
+        # for i in range(0, ...):x
             # Create BoundingBox object and append to list
             # result, box = ...
 
         # Remove this when done
-        raise NotImplementedError
+        #raise NotImplementedError
```
maxlou05, line 118: No need to comment out uneeded code, git will save history for you. You may just delete it.

maxlou05 (commented):
Reviewed

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -64,17 +63,44 @@ def run(self,
...
-        # Remove this when done
-        raise NotImplementedError
-
+        if (self.step == 0):
+            command = commands.Command.create_set_relative_destination_command(self.waypoint.location_x - report.position.location_x, self.waypoint.location_y - report.position.location_y)
+            self.step = 1
+        elif (self.step == 1 and report.status == drone_status.DroneStatus.HALTED):
+            if (getDist(report.position.location_x, report.position.location_y, self.waypoint.location_x, self.waypoint.location_y) <= self.acceptance_radius):
+                self.closest_pad = landing_pad_locations[0]
+                for landing_pad in landing_pad_locations:
```
maxlou05, line 80: Since you set the closest pad to the first element of the list, you no longer need to compare it with the first element of the list. You can skip it with list splicing.

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,33 +85,36 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+        predictions = self.__model.predict(source=image, conf=0.7, device=self.__DEVICE, verbose=True)
 
         # Get the Result object
-        prediction = ...
+        prediction = predictions[0]
 
         # Plot the annotated image from the Result object
         # Include the confidence value
-        image_annotated = ...
+        image_annotated = prediction.plot(conf=0.7)
```
maxlou05, line 95: It should just be `True` to include the confidence value (like it'll just write a number on the output image)

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -64,17 +63,28 @@ def run(self,
...
         # ============
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
-        # Do something based on the report and the state of this class...
+        #Send destination command
+        if (self.on_route == False):
+            command = commands.Command.create_set_relative_destination_command(self.waypoint.location_x - report.position.location_x, self.waypoint.location_y - report.position.location_y)
+            self.on_route = True
+        elif (getDist(report.position.location_x, report.position.location_y, report.destination.location_x, report.destination.location_y) <= self.acceptance_radius):
```
maxlou05, line 76: You may want to check if the drone is halted here, or else it may try to land at the same time the drone is moving towards the waypoint (it is still trying to move to the waypoint when it gets close to the waypoint).

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -64,17 +63,28 @@ def run(self,
...
+
+        # Do something based on the report and the state of this class...
 
         # ============
         # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
         # ============
 
         return command
+
+def getDist(x1, y1, x2, y2):
```
maxlou05, line 89: Using Manhattan distance is unideal in our case, since the drone doesn't have to only move in grid lines. Rather, use Euclidean distance but think about how the fact x^2<y^2 => x<y can be used here (for positive real numbers)

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -64,17 +63,44 @@ def run(self,
...
+        
+        
         # ============
         # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
         # ============
-
         return command
+
+
+def getDist(x1, y1, x2, y2):
```
maxlou05, line 105: Same as above, please use euclidean distance

maxlou05 (commented):
Reviewed

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -64,17 +63,28 @@ def run(self,
...
         # ============
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
-        # Do something based on the report and the state of this class...
+        #Send destination command
+        if (self.on_route == False):
+            command = commands.Command.create_set_relative_destination_command(self.waypoint.location_x - report.position.location_x, self.waypoint.location_y - report.position.location_y)
+            self.on_route = True
+        elif (getDist(report.position.location_x, report.position.location_y, report.destination.location_x, report.destination.location_y) <= self.acceptance_radius and report.status == drone_status.DroneStatus.HALTED):
```
maxlou05, line 76: Nice job not using the square root! Although you should also square the acceptance radius since you are comparing two squared distances

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -64,17 +63,44 @@ def run(self,
...
         # Do something based on the report and the state of this class...
-
-        # Remove this when done
-        raise NotImplementedError
-
+        if (self.step == 0):
+            command = commands.Command.create_set_relative_destination_command(self.waypoint.location_x - report.position.location_x, self.waypoint.location_y - report.position.location_y)
+            self.step = 1
+        elif (self.step == 1 and report.status == drone_status.DroneStatus.HALTED):
+            if (getDist(report.position.location_x, report.position.location_y, self.waypoint.location_x, self.waypoint.location_y) <= self.acceptance_radius):
```
maxlou05, line 78: acceptance radius squared

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -64,17 +63,44 @@ def run(self,
...
+                self.closest_pad = landing_pad_locations[0]
+                for landing_pad in landing_pad_locations[1:]:
+                    if (getDist(landing_pad.location_x, landing_pad.location_y, report.position.location_x, report.position.location_y) < getDist(self.closest_pad.location_x, self.closest_pad.location_y, report.position.location_x, report.position.location_y)):
+                        self.closest_pad = landing_pad
+                command = commands.Command.create_set_relative_destination_command(self.closest_pad.location_x - report.position.location_x, self.closest_pad.location_y - report.position.location_y)
+                self.step = 2  
+            else:
+                self.step = 0
+        elif (self.step == 2 and report.status == drone_status.DroneStatus.HALTED):
+            if (getDist(report.position.location_x, report.position.location_y, self.closest_pad.location_x, self.closest_pad.location_y) <= self.acceptance_radius):
```
maxlou05, line 88: acceptance radius squared

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,33 +85,36 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # move the copy into CPU space,
         # and convert to a numpy array
-        boxes_cpu = ...
+        boxes_cpu = boxes_xyxy.detach().cpu().numpy()
 
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
+        for i in range(boxes_cpu.shape[0]):
+            result, box = bounding_box.BoundingBox.create(bounds=np.array(boxes_cpu[i]))
+            if (result == True):
```
maxlou05, line 109: This works, although in python you don't have to use brackets, you can just do `if result` because without the `== True`, it will return True whenever the object is not None or if it is True

maxlou05 (commented):
Reviewed

maxlou05 (commented):
Approved!

PR #71, opened 2024-06-30
Ajai's Pull Request 

AJ-C22:


modules/bootcamp/test.py
```diff
@@ -0,0 +1,9 @@
+import sys
+import os
+
+# Check if PYTHONPATH is set
+if 'PYTHONPATH' in os.environ:
+    python_path = os.environ['PYTHONPATH']
+    print("PYTHONPATH variable:", python_path)
+else:
+    print("PYTHONPATH variable is not set.")
```
TongguangZhang, line 9: you shouldn't commit these changes since they're for testing the python path and aren't related to the tasks

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,35 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         bounding_boxes = []
         # Hint: .shape gets the dimensions of the numpy array
-        # for i in range(0, ...):
-            # Create BoundingBox object and append to list
-            # result, box = ...
+        for i in range(boxes_cpu.shape[0]):
+            #Create BoundingBox object and append to list
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+            if (result == True):
+                bounding_boxes.append(box)
```
TongguangZhang, line 113: instead of skipping failed box creations, we should return an empty list of bounding boxes instead. In statistics, when we get invalid data, we delete the entire data point instead of part of a data point. In this case a datapoint is 1 image, and the bounding boxes on that image are the parts of the data point, so we would delete the whole image instead of just the invalid boxes.

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,35 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
         # Hint: .shape gets the dimensions of the numpy array
-        # for i in range(0, ...):
-            # Create BoundingBox object and append to list
-            # result, box = ...
+        for i in range(boxes_cpu.shape[0]):
+            #Create BoundingBox object and append to list
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+            if (result == True):
```
TongguangZhang, line 112: avoid unnecessary brackets and extra code, `if result` (or for the comment above, `if not result`) will work here

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,13 +66,56 @@ def run(self,
...
+        if report.status == drone_status.DroneStatus.HALTED:
+            if self.relative_dist_sq(report.position, self.waypoint) > self.acceptance_radius_sq:
+                command = commands.Command.create_set_relative_destination_command(
+                    self.waypoint.location_x - report.position.location_x,
+                    self.waypoint.location_y - report.position.location_y,
+                )
+        elif report.status == drone_status.DroneStatus.MOVING:
+            if self.relative_dist_sq(report.position, self.waypoint) < self.acceptance_radius_sq:
+                command = commands.Command.create_halt_command()
+                self.waypoint_status = True
```
TongguangZhang, line 107: what if the drone halts inside the acceptance radius before this function is called? we would never set `self.waypoint_status = True` and end up in an infinite loop

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,13 +66,56 @@ def run(self,
...
         # ============
         # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
         # ============
 
         return command
+
+    def handle_waypoint_reached(self, report, landing_pad_locations):
+        command = commands.Command.create_null_command()
+        if report.status == drone_status.DroneStatus.HALTED:
+            self.nearest_landing_pad = self.find_nearest_landing_pad(report.position, landing_pad_locations)
```
TongguangZhang, line 83: this searches for the landing pad again after it halts at the landing pad - we should avoid duplicate calculations

TongguangZhang (commented):
reviewed

TongguangZhang (commented):
lgtm

PR #72, opened 2024-07-03
Submission1 7/03

TOMWANGZZ1236:
Submission1

### DIFF
--- a/modules/bootcamp/decision_simple_waypoint.py
+++ b/modules/bootcamp/decision_simple_waypoint.py
@@ -7,6 +7,7 @@ Travel to designated waypoint.
 # pylint: disable=unused-import
 
 
+import math
 from .. import commands
 from .. import drone_report
 from .. import drone_status
@@ -20,10 +21,22 @@ from ..private.decision import base_decision
 
 # All logic around the run() method
 # pylint: disable-next=too-few-public-methods
+def calculate_distance(loc1: location.Location, loc2: location.Location) -> float:
+    return math.sqrt((loc1.location_x - loc2.location_x) ** 2 + (loc1.location_y - loc2.location_y) ** 2)
+
+def calculate_direction(from_loc: location.Location, to_loc: location.Location) -> location.Location:
+        """
+        Calculate the direction vector from current location to waypoint.
+        """
+        direction_x = to_loc.location_x - from_loc.location_x
+        direction_y = to_loc.location_y - from_loc.location_y
+        return location.Location(location_x=direction_x, location_y=direction_y)
+
 class DecisionSimpleWaypoint(base_decision.BaseDecision):
     """
     Travel to the designed waypoint.
     """
+    
     def __init__(self, waypoint: location.Location, acceptance_radius: float):
         """
         Initialize all persistent variables here with self.
@@ -32,49 +45,79 @@ class DecisionSimpleWaypoint(base_decision.BaseDecision):
         print("Waypoint: " + str(waypoint))
 
         self.acceptance_radius = acceptance_radius
-
-        # ============
-        # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
-        # ============
-
-        # Add your own
-
-        # ============
-        # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
-        # ============
-
+        self.has_started_journey = False
     def run(self,
             report: drone_report.DroneReport,
             landing_pad_locations: "list[location.Location]") -> commands.Command:
-        """
-        Make the drone fly to the waypoint.
+        current_location = report.position
+        distance = calculate_distance(current_location, self.waypoint)
+        fly_direction = calculate_direction(current_location, self.waypoint)
+        if report.status == drone_status.DroneStatus.HALTED:
+            if distance <= self.acceptance_radius:
+                print("Arrived, Prepare to land")
+                self.has_started_journey = False
+                return commands.Command.create_land_command()
+            elif not self.has_started_journey:
+                print("Haven't started yet, Prepare to start journey")
+                self.has_started_journey = True
+                return commands.Command.create_set_relative_destination_command(fly_direction.location_x, fly_direction.location_y)
+            else:
+                print('Unexpected halted situation, resume movement')
+                return commands.Command.create_set_relative_destination_command(fly_direction.location_x, fly_direction.location_y)
+            
+        if report.status == drone_status.DroneStatus.MOVING:
+            print('Moving towards the destination')
+            return commands.Command.create_null_command()
+                
+        return commands.Command.create_null_command()   
 
-        You are allowed to create as many helper methods as you want,
-        as long as you do not change the __init__() and run() signatures.
+     
 
-        This method will be called in an infinite loop, something like this:
 
-        ```py
-        while True:
-            report, landing_pad_locations = get_input()
-            command = Decision.run(report, landing_pad_locations)
-            put_output(command)
-        ```
-        """
-        # Default command
-        command = commands.Command.create_null_command()
 
-        # ============
-        # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
-        # ============
 
-        # Do something based on the report and the state of this class...
 
-        # Remove this when done
-        raise NotImplementedError
 
-        # ============
-        # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
-        # ============
 
-        return command
+
+
+
+
+
+
+
+
+
+
+    # def run(self,
+    #         report: drone_report.DroneReport,
+    #         landing_pad_locations: "list[location.Location]") -> commands.Command:
+    #     current_location = report.position
+    #     distance = calculate_distance(current_location, self.waypoint)
+    #     # Default command
+    #     command = commands.Command.create_null_command()
+    #     if report.status == drone_status.DroneStatus.HALTED:
+    #         if distance <= self.acceptance_radius:
+    #             print("Waypoint reached. Initiating landing.")
+    #             self.has_started_journey = False  # Reset for potential future commands
+    #             return commands.Command.create_land_command()
+            
+    #         elif not self.has_started_journey:
+    #             # Drone is halted at the initial position, not yet moved.
+    #             print("Starting journey towards the waypoint.")
+    #             direction_to_move = calculate_direction(current_location, self.waypoint)
+    #             self.has_started_journey = True
+    #             return commands.Command.create_set_relative_destination_command(direction_to_move.location_x, direction_to_move.location_y)
+    #         elif self.has_started_journey:
+    #             # Unexpected halt, possibly check if it needs to resume course.
+    #             print("Unexpected halt detected, resuming course.")
+    #             direction_to_move = calculate_direction(current_location, self.waypoint)
+    #             return commands.Command.create_set_relative_destination_command(direction_to_move.location_x, direction_to_move.location_y)
+            
+    #     # If flying and not within radius, continue to fly towards the waypoint.
+    #     if report.status == drone_status.DroneStatus.MOVING:
+    #         print("In flight towards waypoint, continuing as planned.")
+    #         return commands.Command.create_null_command()
+
+    #     return commands.Command.create_null_command()
+
--- a/modules/bootcamp/decision_waypoint_landing_pads.py
+++ b/modules/bootcamp/decision_waypoint_landing_pads.py
@@ -7,6 +7,7 @@ Travel to designated waypoint and then land at a nearby landing pad.
 # pylint: disable=unused-import
 
 
+import math
 from .. import commands
 from .. import drone_report
 from .. import drone_status
@@ -20,6 +21,29 @@ from ..private.decision import base_decision
 
 # All logic around the run() method
 # pylint: disable-next=too-few-public-methods
+def calculate_closest_location(start_loc: location.Location, loc_list:"list[location.Location]") -> location.Location:
+    shortest = float('inf')
+    
+    for landing_loc in loc_list:
+        if calculate_distance(start_loc, landing_loc) < shortest:
+            shortest = calculate_distance(start_loc, landing_loc)
+            shortest_location = landing_loc
+    
+    return shortest_location
+
+
+
+def calculate_distance(loc1: location.Location, loc2: location.Location) -> float:
+    return math.sqrt((loc1.location_x - loc2.location_x) ** 2 + (loc1.location_y - loc2.location_y) ** 2)
+
+def calculate_direction(from_loc: location.Location, to_loc: location.Location) -> location.Location:
+        """
+        Calculate the direction vector from current location to waypoint.
+        """
+        direction_x = to_loc.location_x - from_loc.location_x
+        direction_y = to_loc.location_y - from_loc.location_y
+        return location.Location(location_x=direction_x, location_y=direction_y)
+
 class DecisionWaypointLandingPads(base_decision.BaseDecision):
     """
     Travel to the designed waypoint and then land at the nearest landing pad.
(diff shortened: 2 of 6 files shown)


modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -7,6 +7,7 @@
 # pylint: disable=unused-import
 
 
+import math
```
TongguangZhang, line 10: do not import additional libraries, this bootcamp can be completed without the math library

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -32,49 +45,79 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
...
 
-        # ============
-        # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
-        # ============
 
-        # Add your own
 
-        # ============
-        # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
-        # ============
```
TongguangZhang, line 44: do not modify the comment guards

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -32,49 +45,79 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
...
-        as long as you do not change the __init__() and run() signatures.
 
-        This method will be called in an infinite loop, something like this:
 
-        ```py
-        while True:
-            report, landing_pad_locations = get_input()
-            command = Decision.run(report, landing_pad_locations)
-            put_output(command)
-        ```
```
TongguangZhang, line 62: do not modify the function signature

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -32,49 +45,79 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
         print("Waypoint: " + str(waypoint))
 
         self.acceptance_radius = acceptance_radius
+        self.has_started_journey = False
+    def run(self,
+            report: drone_report.DroneReport,
+            landing_pad_locations: "list[location.Location]") -> commands.Command:
+        current_location = report.position
```
TongguangZhang, line 52: don't rename variables - use report.position throughout so that it's easy to keep track of (instead of remembering 2 variable names we only need to remember 1)

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -32,49 +45,79 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
...
+                print("Haven't started yet, Prepare to start journey")
+                self.has_started_journey = True
+                return commands.Command.create_set_relative_destination_command(fly_direction.location_x, fly_direction.location_y)
+            else:
+                print('Unexpected halted situation, resume movement')
+                return commands.Command.create_set_relative_destination_command(fly_direction.location_x, fly_direction.location_y)
+            
+        if report.status == drone_status.DroneStatus.MOVING:
+            print('Moving towards the destination')
+            return commands.Command.create_null_command()
```
TongguangZhang, line 70: this case is redundant as it is covered by the base case (both return null commands)
however, the case of the drone moving is important to consider - what if the drone is within the acceptance radius of the waypoint and still moving? should we let it keep moving?

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -32,49 +45,79 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
...
+            
+    #     # If flying and not within radius, continue to fly towards the waypoint.
+    #     if report.status == drone_status.DroneStatus.MOVING:
+    #         print("In flight towards waypoint, continuing as planned.")
+    #         return commands.Command.create_null_command()
 
-        # ============
-        # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
-        # ============
+    #     return commands.Command.create_null_command()
```
TongguangZhang, line 122: remove commented out or "dead" code - this is code that is never going to run, if we need it, we can go back in the commit history for it, but leaving it commented out clutters up the code and is unnecessary
to clarify, annotative comments are good, commented out code is bad

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -7,6 +7,7 @@
 # pylint: disable=unused-import
 
 
+import math
```
TongguangZhang, line 10: do not import math

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -20,10 +21,22 @@
...
+def calculate_distance(loc1: location.Location, loc2: location.Location) -> float:
+    return math.sqrt((loc1.location_x - loc2.location_x) ** 2 + (loc1.location_y - loc2.location_y) ** 2)
+
+def calculate_direction(from_loc: location.Location, to_loc: location.Location) -> location.Location:
+        """
+        Calculate the direction vector from current location to waypoint.
+        """
+        direction_x = to_loc.location_x - from_loc.location_x
+        direction_y = to_loc.location_y - from_loc.location_y
+        return location.Location(location_x=direction_x, location_y=direction_y)
```
TongguangZhang, line 33: define these as static functions in the class - the DecisionSimpleWaypoint class is the only thing that uses them, and since it does not take in instance or class variables (self or cls), we make them static

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -20,6 +21,29 @@
...
+def calculate_distance(loc1: location.Location, loc2: location.Location) -> float:
+    return math.sqrt((loc1.location_x - loc2.location_x) ** 2 + (loc1.location_y - loc2.location_y) ** 2)
+
+def calculate_direction(from_loc: location.Location, to_loc: location.Location) -> location.Location:
+        """
+        Calculate the direction vector from current location to waypoint.
+        """
+        direction_x = to_loc.location_x - from_loc.location_x
+        direction_y = to_loc.location_y - from_loc.location_y
+        return location.Location(location_x=direction_x, location_y=direction_y)
```
TongguangZhang, line 45: define these as static functions in the class - the DecisionWaypointLandingPads class is the only thing that uses them, and since it does not take in instance or class variables (self or cls), we make them static

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -30,51 +54,45 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
...
-
-        This method will be called in an infinite loop, something like this:
-
-        ```py
-        while True:
-            report, landing_pad_locations = get_input()
-            command = Decision.run(report, landing_pad_locations)
-            put_output(command)
-        ```
-        """
```
TongguangZhang, line 63: do not remove the docstring

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -30,51 +54,45 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
...
-
-        # Remove this when done
-        raise NotImplementedError
-
-        # ============
-        # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
-        # ============
-
-        return command
+        current_location = report.position
```
TongguangZhang, line 67: do not rename this variable, use report.position throughout

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -30,51 +54,45 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
...
+                    print("Haven't started yet, Prepare to start journey")
+                    self.has_started_journey = True
+                    return commands.Command.create_set_relative_destination_command(fly_direction.location_x, fly_direction.location_y)
+                else:
+                    print('Unexpected halted situation, resume movement')
+                    return commands.Command.create_set_relative_destination_command(fly_direction.location_x, fly_direction.location_y)
+       
+        if report.status == drone_status.DroneStatus.MOVING:
+                    print('Moving towards the destination')
+                    return commands.Command.create_null_command()
```
TongguangZhang, line 94: this case is redundant as it is covered by the base case (both return null commands)
however, the case of the drone moving is important to consider - what if the drone is within the acceptance radius of the waypoint and still moving? should we let it keep moving?

modules/bootcamp/detect_landing_pad.py
```diff
@@ -8,6 +8,7 @@
 import numpy as np
 import torch
 import ultralytics
+import cv2
```
TongguangZhang, line 11: do not import additional libraries, this bootcamp can be completed without cv2

modules/bootcamp/detect_landing_pad.py
```diff
@@ -79,30 +80,40 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+        predictions = self.__model.predict([image]) 
 
         # Get the Result object
-        prediction = ...
+        prediction = predictions[0]
 
         # Plot the annotated image from the Result object
         # Include the confidence value
-        image_annotated = ...
+        image_annotated = np.copy(image)
```
TongguangZhang, line 97: we want the annotated image, which includes the bounding boxes, not the copy of the original image

modules/bootcamp/detect_landing_pad.py
```diff
@@ -79,30 +80,40 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Get the xyxy boxes list from the Boxes object in the Result object
-        boxes_xyxy = ...
+        boxes_xyxy = prediction.boxes.xyxy
 
         # Detach the xyxy boxes to make a copy,
         # move the copy into CPU space,
         # and convert to a numpy array
-        boxes_cpu = ...
-
+        boxes_cpu = boxes_xyxy.cpu().numpy()
```
TongguangZhang, line 105: make sure to call detach() on this - this removes it from the computational graph
this turns off the gradients, which are by default on for model training purposes (gradients are used in the back propagation step)

modules/bootcamp/detect_landing_pad.py
```diff
@@ -79,30 +80,40 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+        bounding_boxes = []
+        for box in boxes_cpu:
+            x1, y1, x2, y2 = map(int, box)
+            cv2.rectangle(image_annotated, (x1, y1), (x2, y2), (0, 255, 0), 2)
+            label = f"{x1}, {y1}, {x2}, {y2}"
+            cv2.putText(image_annotated, label, (x1, y1 - 10), cv2.FONT_HERSHEY_SIMPLEX, 0.5, (255, 0, 0), 2)
```
TongguangZhang, line 111: you don't have to do this manually, ultralytics has the functionality for this

modules/bootcamp/detect_landing_pad.py
```diff
@@ -79,30 +80,40 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+            result_box = bounding_box.BoundingBox.create(np.array([x1, y1, x2, y2]))[1]
```
TongguangZhang, line 112: you can just pass in the box since it's already an np array of 4 ints

modules/bootcamp/detect_landing_pad.py
```diff
@@ -79,30 +80,40 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+            if result_box is not None:
```
TongguangZhang, line 113: instead of checking if the box is none, use the result flag returned in the create method:
```
result, bounding_box = bounding_box.BoundingBox.create(box)

if result:
    # do something
```

modules/bootcamp/detect_landing_pad.py
```diff
@@ -79,30 +80,40 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+                bounding_boxes.append(result_box)
```
TongguangZhang, line 114: instead of skipping failed box creations, we should return an empty list of bounding boxes instead. In statistics, when we get invalid data, we delete the entire data point instead of part of a data point. In this case a datapoint is 1 image, and the bounding boxes on that image are the parts of the data point, so we would delete the whole image instead of just the invalid boxes.

modules/bootcamp/detect_landing_pad.py
```diff
@@ -79,30 +80,40 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+
+        return bounding_boxes, image_annotated
```
TongguangZhang, line 116: remove the code after this - code after the return is also "dead" code - the assignment of `bounding_boxes = []` is never run

TongguangZhang (commented):
reviewed

modules/bootcamp/detect_landing_pad.py
TOMWANGZZ1236: I’m thinking of returning ([], image_annotated) if the box creation fails. It handles the case of failure by returning an empty list for bounding boxes ([]), but what should I return for image_annotated? If the creation fails, should I just return the original image?

modules/bootcamp/decision_waypoint_landing_pads.py
TOMWANGZZ1236: My thoughts are that we should issue a halt command when the drone is located within the acceptance radius of the waypoint since it needs to be prepared for landing if it is already within the acceptance radius, and the command of landing needs it be in the state of halted first. However, I also observed that it is automatically issued a halt command when it reached the waypoint. Can you provide some clarifications?

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -20,32 +21,43 @@
...
 class DecisionSimpleWaypoint(base_decision.BaseDecision):
     """
     Travel to the designed waypoint.
     """
-    def __init__(self, waypoint: location.Location, acceptance_radius: float):
+    @staticmethod
+    def calculate_distance(loc1: location.Location, loc2: location.Location) -> float:
+        return ((loc1.location_x - loc2.location_x) ** 2 + (loc1.location_y - loc2.location_y) ** 2) ** 0.5
+
+    def calculate_direction(from_loc: location.Location, to_loc: location.Location) -> location.Location:
```
TongguangZhang, line 33: this should also be a staticmethod

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -67,14 +79,37 @@ def run(self,
         # ============
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
-
+        distance = DecisionSimpleWaypoint.calculate_distance(report.position, self.waypoint)
+        fly_direction = DecisionSimpleWaypoint.calculate_direction(report.position, self.waypoint)
```
TongguangZhang, line 83: keep variables in the lowest scope possible - since this isn't used outside the report.status = HALTED case, only declare this variable in that scope

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -20,7 +20,31 @@
 
 # All logic around the run() method
 # pylint: disable-next=too-few-public-methods
+
+
 class DecisionWaypointLandingPads(base_decision.BaseDecision):
+    staticmethod
```
TongguangZhang, line 26: `@staticmethod` - otherwise this line should actually result in an error

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -20,7 +20,31 @@
...
+    def calculate_closest_location(start_loc: location.Location, loc_list:"list[location.Location]") -> location.Location:
+        shortest = float('inf')
+        
+        for landing_loc in loc_list:
+            if DecisionWaypointLandingPads.calculate_distance(start_loc, landing_loc) < shortest:
+                shortest = DecisionWaypointLandingPads.calculate_distance(start_loc, landing_loc)
+                shortest_location = landing_loc
+        
+        return shortest_location
```
TongguangZhang, line 35: what if the list is empty? this would result in an undeclared variable error - make sure to declare shortest_location outside the for loop (maybe initialize it to None

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -20,7 +20,31 @@
 
 # All logic around the run() method
 # pylint: disable-next=too-few-public-methods
+
+
 class DecisionWaypointLandingPads(base_decision.BaseDecision):
+    staticmethod
+    def calculate_closest_location(start_loc: location.Location, loc_list:"list[location.Location]") -> location.Location:
```
TongguangZhang, line 27: make this a static method

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -20,7 +20,31 @@
...
+        shortest = float('inf')
+        
+        for landing_loc in loc_list:
+            if DecisionWaypointLandingPads.calculate_distance(start_loc, landing_loc) < shortest:
+                shortest = DecisionWaypointLandingPads.calculate_distance(start_loc, landing_loc)
+                shortest_location = landing_loc
+        
+        return shortest_location
+
+    def calculate_distance(loc1: location.Location, loc2: location.Location) -> float:
```
TongguangZhang, line 37: make this a static method

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -20,7 +20,31 @@
...
+        return ((loc1.location_x - loc2.location_x) ** 2 + (loc1.location_y - loc2.location_y) ** 2)**0.5
+
+    def calculate_direction(from_loc: location.Location, to_loc: location.Location) -> location.Location:
```
TongguangZhang, line 40: make this a static method

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -20,7 +20,31 @@
...
+            """
+            Calculate the direction vector from current location to waypoint.
+            """
+            direction_x = to_loc.location_x - from_loc.location_x
+            direction_y = to_loc.location_y - from_loc.location_y
+            return location.Location(location_x=direction_x, location_y=direction_y)
+
```
TongguangZhang, line 47: by convention, __init__ should be the first function in the class, and all functions should go below the docstring

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -38,14 +62,17 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
...
 
         # ============
         # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
         # ============
 
     def run(self,
-            report: drone_report.DroneReport,
-            landing_pad_locations: "list[location.Location]") -> commands.Command:
+        report: drone_report.DroneReport,
+        landing_pad_locations: "list[location.Location]") -> commands.Command:
```
TongguangZhang, line 75: try to keep your changes inside the comment guards (this spacing is against the style convention)

modules/bootcamp/detect_landing_pad.py
```diff
@@ -65,6 +66,68 @@ def __init__(self, class_private_create_key, model: ultralytics.YOLO):
...
+    #     # for i in range(0, ...):
+    #         # Create BoundingBox object and append to list
+    #         # result, box = ...
+
+    #     # Remove this when done
+        
+
+    #     # ============
+    #     # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
+    #     # ============
```
TongguangZhang, line 130: remove commented out code

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,33 +149,39 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+        boxes_cpu = boxes_xyxy.cpu().numpy()
 
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
+        for box in boxes_cpu:         
+            result, result_box = bounding_box.BoundingBox.create(box)
+            if result:
+                bounding_boxes.append(result_box)
+            else:
+                return ([], image_annotated)
```
TongguangZhang, line 176: `return [], image_annotated`

TongguangZhang, line 176: since we return on the false case, this can just be:
```
if not result:
    return [], image_annotated
bounding_boxes.append(result_box)
```
this simplified the code a little

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,33 +149,39 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+
+        return (bounding_boxes, image_annotated)
```
TongguangZhang, line 178: `return bounding_boxes, image_annotateded`
avoid unnecessary brackets

TongguangZhang (commented):
reviewed

TongguangZhang (commented):
lgtm

PR #73, opened 2024-07-12
Jacky Men

jackeyemean:


modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -37,41 +37,36 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
...
-
-        This method will be called in an infinite loop, something like this:
 
-        ```py
-        while True:
-            report, landing_pad_locations = get_input()
-            command = Decision.run(report, landing_pad_locations)
-            put_output(command)
-        ```
-        """
```
maxlou05, line 63: Although you don't need this for the bootcamp, we require writing docstrings for our functions and classes (just an FYI, you don't need to change anything here).

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -37,41 +37,36 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
...
-        # Remove this when done
-        raise NotImplementedError
+        
+        if report.status == drone_status.DroneStatus.HALTED: 
+
+            x_value_diff = self.waypoint.location_x - report.position.location_x
+            y_value_diff = self.waypoint.location_y - report.position.location_y
+
+            # Distance formula
+            distance_from_waypoint = (x_value_diff ** 2 + y_value_diff ** 2) ** 1/2
```
maxlou05, line 61: If you take into account order of operations, the program wouldn't work as intended here. Additionally, think about how to compare distances without the square root operation.

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,33 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
-        # for i in range(0, ...):
-            # Create BoundingBox object and append to list
-            # result, box = ...
-
-        # Remove this when done
-        raise NotImplementedError
+        # .shape gets the dimensions of the numpy array
+        for i in range(0, boxes_cpu.shape[0]):
+            # line below uses tuple unpacking
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
```
maxlou05, line 111: Great usage of tuple unpacking! (You do not need to change anything here)

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,33 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+            if result and box is not None:
```
maxlou05, line 112: Usually in statistics, if one part of the data is invalid (ie a bounding box), then the whole data point is discarded (ie the whole image).

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -46,32 +47,58 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
...
 
-        # Do something based on the report and the state of this class...
-
-        # Remove this when done
-        raise NotImplementedError
+        if report.status == drone_status.DroneStatus.HALTED:
+            # Getting distance to waypoint
+            x_diff = self.waypoint.location_x - report.position.location_x
+            y_diff = self.waypoint.location_y - report.position.location_y
+            distance_to_waypoint = (x_diff ** 2 + y_diff ** 2) ** 0.5
```
maxlou05, line 62: Again, try to avoid using the square root function (just a micro-optimization, which could be impactful on low-power machines such as the drone's processor). (there are also a few other instances of this calculation later on)

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -46,32 +47,58 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
...
+                    for pad in landing_pad_locations:
+                        # Getting distance to landing pad
+                        pad_x_diff = pad.location_x - report.position.location_x
+                        pad_y_diff = pad.location_y - report.position.location_y
+                        pad_distance = (pad_x_diff ** 2 + pad_y_diff ** 2) ** 0.5
+
+                        # Comparing to closest landing pad so far
+                        if pad_distance < smallest_pad_distance:
+                            smallest_pad_distance = pad_distance
+                            self.landing_pad = pad
```
maxlou05, line 81: Although in this bootcamp you are guaranteed to find at least 1 landing pad, what would happen to the code if there were none? (You do not need to change anything here, just something to think about)

maxlou05 (commented):
Reviewed.

maxlou05 (commented):
Approved.

PR #75, opened 2024-07-28
Bootcamp PR

LithiumEngineer:


modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -70,9 +75,37 @@ def run(self,
...
+            if self.at_location(report.position, self.waypoint): 
+                command = commands.Command.create_halt_command()
+                self.at_waypoint = True
+        elif report.status == drone_status.DroneStatus.HALTED or self.begin_landing: 
+            if not self.at_waypoint: 
+                command = commands.Command.create_set_relative_destination_command(
+                    self.waypoint.location_x - report.position.location_x, 
+                    self.waypoint.location_y - report.position.location_y
+                )
+                self.at_waypoint = True
```
maxlou05, line 88: I think there's no need to set at_waypoint to true here since you will set it to true once you get there eventually. There's no need to set it early, just in case something wrong happens.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -70,9 +75,37 @@ def run(self,
...
+                self.begin_landing = True
+                res = None
+                mn = float('inf')
+                for i in landing_pad_locations: 
+                    dx = i.location_x - report.position.location_x
+                    dy = i.location_y - report.position.location_y
+                    dist = dx ** 2 + dy ** 2
+                    if dist < mn: 
+                        mn = dist
+                        res = i
```
maxlou05, line 100: This code is fine, but we encourage using more descriptive variable names, even if it's longer to type out. For example, instead of res, we can have nearest_landing_pad, and mn as min_dist or dist_to_nearest_pad

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -70,9 +75,37 @@ def run(self,
...
         # Do something based on the report and the state of this class...
 
-        # Remove this when done
-        raise NotImplementedError
-
+        if report.status == drone_status.DroneStatus.MOVING and not self.begin_landing:
+            if self.at_location(report.position, self.waypoint): 
+                command = commands.Command.create_halt_command()
+                self.at_waypoint = True
+        elif report.status == drone_status.DroneStatus.HALTED or self.begin_landing:
```
maxlou05, line 82: Although I do believe that this logic works, it may be confusing to readers because they have to deduce by themselves that no other landing pad can get closer than the designated "closest landing pad" while en route to this "closest landing pad" from the waypoint. It is probably better to split these up into 2 different cases rather than trying to handle them at the same time. (Note this attempts to handle 3 of 4 possible outcomes)

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -31,14 +31,17 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
...
 
         # ============
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
         # Add your own
-
+        self.at_waypoint = False
+        self.landing_sent = False
+    def at_location(self, cur: location.Location, dest: location.Location):
```
maxlou05, line 43: Great job recognizing that you can compare distances without using the square root operation! (you do not need to change anything here)

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -31,14 +31,17 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
         self.waypoint = waypoint
         print("Waypoint: " + str(waypoint))
 
-        self.acceptance_radius = acceptance_radius
+        self.acceptance_radius = acceptance_radius ** 2
```
maxlou05, line 34: Great job using the acceptance radius! (you do not need to change anything here)

maxlou05 (commented):
Reviewed

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -70,9 +75,35 @@ def run(self,
...
+            print("now begin landing!")
+            nearest_landing_pad = None
+            min_dist = float('inf')
+            for i in landing_pad_locations: 
+                dx = i.location_x - report.position.location_x
+                dy = i.location_y - report.position.location_y
+                dist = dx ** 2 + dy ** 2
+                if dist < min_dist: 
+                    min_dist = dist
+                    nearest_landing_pad = i
```
maxlou05, line 88: I think it is better to not re-calculate the nearest landing pad unnecessarily all the time. It is better to move this to the section when you have just arrived at the waypoint, since you will never get closer to a different landing pad while approaching this one. Additionally, i is not a great variable name (this is not quite the same as a traditional c++ for loop), please consider changing it to something like landing_pad.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -70,9 +75,35 @@ def run(self,
...
+                else: 
+                    command = commands.Command.create_set_relative_destination_command(
+                        nearest_landing_pad.location_x - report.position.location_x, 
+                        nearest_landing_pad.location_y - report.position.location_y
+                    )
+        elif report.status == drone_status.DroneStatus.MOVING:
+            if self.at_location(report.position, self.waypoint): 
+                command = commands.Command.create_halt_command()
+                self.begin_landing = True
+        elif report.status == drone_status.DroneStatus.HALTED:
```
maxlou05, line 101: Note that your drone will get stuck in this condition forever if it ever comes in here. It it's halted at the waypoint and begin_landing is not true, you will be stuck here forever.

maxlou05 (commented):
Reviewed

PR #78, opened 2024-09-01
Autonomy Bootcamp - Krish

Krish-310:


modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -71,7 +73,65 @@ def run(self,
...
+        closest_landing_pad_distance = float('inf')
+        
+        if self.reached_waypoint:
+            # If reached waypoint, then find nearest landing pad and move towards it
+            for landing_pad in landing_pad_locations:
+
+                relative_x = landing_pad.location_x - report.position.location_x
+                relative_y = landing_pad.location_y - report.position.location_y
+                distance = get_distance(relative_x, relative_y)
+                # If distance < acceptance radius, then the current position (waypoint) is a landing pad, so land
```
AaronWang04, line 93: this is not true. you need to get into range of a landing pad zone (distance < accepted radius) and then select the best landing pad to land on

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +88,41 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+        num_boxes, _ = boxes_cpu.shape
         # for i in range(0, ...):
             # Create BoundingBox object and append to list
             # result, box = ...
-
-        # Remove this when done
-        raise NotImplementedError
+        for i in range(num_boxes):
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+            if result:
```
AaronWang04, line 119: generally if result is none then the whole data should be thrown out to indicate error with the batch, tho this implementation is understandable as well

modules/bootcamp/decision_waypoint_landing_pads.py
Krish-310: I'm not sure that I fully understand what the goal of the task here is... 
First, I ensure that the drone has reached the appropriate waypoint. Then, if the waypoint happens to be the within the acceptance radius of a landing pad, I can land it right there (that is what this condition handles). Then, if this is not the case, then we use the closest landing pad location we got and set that as our new destination (waypoint). Then this condition would handle the landing on that landing pad as well.
Am I missing something here?

AaronWang04: yeah this confusion is not unwarranted. think about why specifically we need acceptance radius.

it is because location of the drone is always analog (as reported by the gps). if the drone wants to go to (1,1) and it is at (0.998, 1.001), it is not exactly on the location but it is close enoug within margin of error where we would consider it having reached the location.

the task at hand is essentially for the drone to reach a location where there will be landing pads, then we will call detect_landing_pad which is a cv model that looks for landing pads with the camera (this is abstracted away so u dont really have to worry about it, the result of the model is passed into the run function), and then choose the best landing pad to land from that list

AaronWang04 (approved):
approved

PR #80, opened 2024-09-05
Autonomy Bootcamp - Vilo

Phiruby:


modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,42 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Get the Result object
-        prediction = ...
+        prediction = predictions[0]
+        # print(prediction)
 
         # Plot the annotated image from the Result object
         # Include the confidence value
-        image_annotated = ...
+        # image_annotated = prediction.masks
+        image_annotated = prediction.plot()
```
maxlou05, line 103: Please include the confidence value as requested. (Hint, use conf kwarg)

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,42 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
 
         # Get the xyxy boxes list from the Boxes object in the Result object
-        boxes_xyxy = ...
+        boxes_xyxy = prediction.boxes.xyxy
 
         # Detach the xyxy boxes to make a copy,
         # move the copy into CPU space,
         # and convert to a numpy array
-        boxes_cpu = ...
+        boxes_cpu = np.array(boxes_xyxy.detach().clone()) # <--- first dimension is number of bounding box
```
maxlou05, line 111: Please first copy to CPU space (there should be a cpu()), then you can directly get numpy array, you don't need to manually convert tensor to numpy

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,42 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
         # Hint: .shape gets the dimensions of the numpy array
-        # for i in range(0, ...):
+        for i in range(0, boxes_cpu.shape[0]):
             # Create BoundingBox object and append to list
-            # result, box = ...
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+            if not result or not box:
+                raise Exception("Error creating bounding box! Tried using data: ", boxes_cpu[i])
```
maxlou05, line 120: Although this works, we don't want to raise an exception in this case, which will crash the code. For this bootcamp, just return empty list.

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,42 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
 
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
         # Hint: .shape gets the dimensions of the numpy array
-        # for i in range(0, ...):
+        for i in range(0, boxes_cpu.shape[0]):
             # Create BoundingBox object and append to list
-            # result, box = ...
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+            if not result or not box:
```
maxlou05, line 119: You don't need to check both result and box, result will tell you if it succeded or not.

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -70,8 +87,41 @@ def run(self,
...
+            return command 
+
+        # if waypoint is reached, we land
+        if self.waypoint_reached:
+            print("Landing")
+            self.waypoint_landed = True
+            return commands.Command.create_land_command()
+
+        # checks if the drone is within the acceptance radius
+        if self.distance_to_waypoint(report) <= self.acceptance_radius ** 2:
```
maxlou05, line 104: Great job recognizing that you can compare squared distances to save computation time! (You don't need to change anything here)

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -38,11 +38,28 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
         # ============
 
         # Add your own
+        self.waypoint_reached = False
+        self.waypoint_reached_count = 0
+        self.waypoint_landed = False
```
maxlou05, line 43: I don't think you ended up using these two variables, so you can delete them.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -38,10 +38,61 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
...
+        Calculate the relative displacement between the drone and the landing pad.
+        """
+        return location.Location(landing_pad_location.location_x - drone_location.location_x, landing_pad_location.location_y - drone_location.location_y)
+
+    def get_closest_landing_pad(self, landing_pad_locations: "list[location.Location]") -> location.Location:
+        """
+        Get the closest landing pad to the drone (from the waypoint).
+        """
+        # first assume the closest landing pad is the first one
+        closest_landing_pad = landing_pad_locations[0]
```
maxlou05, line 86: For this bootcamp, this will work. But just a thought: what would happen if the list you are given was empty? (You do not need to change anything here)

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,12 +120,61 @@ def run(self,
...
+        if report.status == drone_status.DroneStatus.HALTED and not self.waypoint_reached:
+            print("Moving to waypoint")
+
+            relative_destination_from_current_position = self.get_relative_displacement(report.position)
+            command = commands.Command.create_set_relative_destination_command(
+                relative_destination_from_current_position.location_x,
+                relative_destination_from_current_position.location_y
+            )
+            # Remove this when done
+            # raise NotImplementedError
```
maxlou05, line 174: You can just remove these comments, as git will store old code in history.

maxlou05 (commented):
Reviewed.

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,40 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
         # Hint: .shape gets the dimensions of the numpy array
-        # for i in range(0, ...):
+        for i in range(0, boxes_cpu.shape[0]):
             # Create BoundingBox object and append to list
-            # result, box = ...
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+            if not result:
+                box = []
```
maxlou05, line 118: You are appending a empty list to an empty list, which is probably not desired.

modules/bootcamp/detect_landing_pad.py
```diff
@@ -86,32 +86,40 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
 
         # Get the xyxy boxes list from the Boxes object in the Result object
-        boxes_xyxy = ...
+        boxes_xyxy = prediction.boxes.xyxy
 
         # Detach the xyxy boxes to make a copy,
         # move the copy into CPU space,
         # and convert to a numpy array
-        boxes_cpu = ...
+        boxes_cpu = boxes_xyxy.cpu().numpy()  # <--- first dimension is number of bounding box
```
maxlou05, line 109: Also detach before doing cpu

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -38,10 +38,61 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float):
         # ============
 
         # Add your own
+        self.waypoint_reached = False
+        self.reached_landing_pad = False
```
maxlou05, line 42: You also probably don't need this, since you never read from it.

maxlou05 (commented):
Reviewed

maxlou05 (commented):
Approved!

PR #81, opened 2024-09-08
Autonomy Bootcamp Submission - Daniel W

wdan30:


modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,7 +68,17 @@ def run(
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
-        # Do something based on the report and the state of this class...
+        print(self.waypoint == report.position)
+
+        if self.waypoint == report.position:
```
AaronWang04, line 73: this will almost never evaluate as true in real scenarios where report.position is obtained via an analog gps device (can you think of why?). hence why we have acceptance radius as a member variable

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +96,35 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # move the copy into CPU space,
         # and convert to a numpy array
-        boxes_cpu = ...
+        boxes_cpu = boxes_xyxy.cpu().numpy()
 
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
+
+        for box in boxes_cpu:
+            bounding_boxes.append(bounding_box.BoundingBox.create(box)[1])
```
AaronWang04, line 120: look at the returns of boundingbox.create function, are there exceptions to catch? if so, you should handle it

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -36,8 +36,30 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
         # ============
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
+        self.landed_at_waypoint = False
 
-        # Add your own
+    def __get_distance(self, position_1: location.Location, position_2: location.Location) -> float:
```
AaronWang04, line 41: good use of type hinting!

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -36,8 +36,30 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
...
+        return (
+            (position_1.location_x - position_2.location_x) ** 2
+            + (position_1.location_y - position_2.location_y) ** 2
+        ) ** 0.5
```
AaronWang04, line 45: ** 0.5 is a computationally expensive operation, can you think of a way to avoid using it? you dont need to make any changes here as implementaiton is perfectly fine but its just something to think about

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,7 +90,21 @@ def run(
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
-        # Do something based on the report and the state of this class...
+        if self.waypoint == report.position:
```
AaronWang04, line 93: same issue here as described in first task

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,7 +90,21 @@ def run(
...
+            if self.landed_at_waypoint:
+                command = commands.Command.create_land_command()
```
AaronWang04, line 95: you can return here to avoid calling subsequent code again

AaronWang04 (commented):
reviewed

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,7 +68,18 @@ def run(
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
-        # Do something based on the report and the state of this class...
+        if (
+            abs(self.waypoint.location_x - report.position.location_x) < 0.1
```
AaronWang04, line 72: intended solution but self.accepted_radius is specified, so you should use that. note that it is radius

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +96,38 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
-        boxes_cpu = ...
+        boxes_cpu = boxes_xyxy.cpu().numpy()
 
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
+
+        for box in boxes_cpu:
+            bounding_box_output = bounding_box.BoundingBox.create(box)
+
+            if bounding_box_output[0]:
```
AaronWang04, line 122: good handling, although the general convention is to disgard the whole image if any of the datapoints was errored. don't have to change anything just smth to think about

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,7 +89,22 @@ def run(
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
-        # Do something based on the report and the state of this class...
+        if (
+            abs(self.waypoint.location_x - report.position.location_x) < 0.1
```
AaronWang04, line 93: same issue as above

AaronWang04 (commented):
almost there!

AaronWang04 (approved):

PR #83, opened 2024-09-09
Aleksa - Autonomy Bootcamp

Cyuber:


modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -38,6 +38,16 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
         # ============
 
         # Add your own
+        self.command_index = 0
+        self.commands = [
+            #commands.Command.create_set_relative_destination_command(30.0, 30.0),
+            #commands.Command.create_set_relative_destination_command(-30.0, 30.0),
```
maxlou05, line 44: You can delete these testing codes. We usually reserve comments for explaining the code, git can save old code if needed.

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -69,6 +79,17 @@ def run(
...
+            command = self.commands[self.command_index]
+            self.command_index += 1
+        
+        elif report.status == drone_status.DroneStatus.HALTED:
+            command = commands.Command.create_land_command()
+
+            self.has_sent_landing_command = True
+
+        if self.has_sent_landing_command:
+            command = commands.Command.create_land_command()
```
maxlou05, line 92: I don't think it is necessary to have two conditions that do the same thing. You should be able to combine these.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -38,6 +38,17 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
         # ============
 
         # Add your own
+        self.command_index = 0
+        self.commands = [
+            #commands.Command.create_set_relative_destination_command(30.0, 30.0),
+            #commands.Command.create_set_relative_destination_command(-30.0, 30.0),
```
maxlou05, line 44: Same, no need to comment out code

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -38,6 +38,16 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
         # ============
 
         # Add your own
+        self.command_index = 0
+        self.commands = [
+            #commands.Command.create_set_relative_destination_command(30.0, 30.0),
+            #commands.Command.create_set_relative_destination_command(-30.0, 30.0),
+            commands.Command.create_set_relative_destination_command(self.waypoint.location_x, self.waypoint.location_y)
```
maxlou05, line 45: Since this is a relative movement command, it depends upon the drone's position as well. You may end up going to a different location than you expected if you only put the waypoint's location.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -38,6 +38,17 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
...
+            commands.Command.create_set_relative_destination_command(self.waypoint.location_x, self.waypoint.location_y)
```
maxlou05, line 45: Same, please use relative distance

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,6 +80,31 @@ def run(
         # ============
 
         # Do something based on the report and the state of this class...
+        if report.status == drone_status.DroneStatus.HALTED and self.command_index < len(self.commands):
+            command = self.commands[self.command_index]
+            self.command_index += 1
+        
+        elif report.status == drone_status.DroneStatus.HALTED and not self.has_found_landing_pad:
+            dist_min = 30000
```
maxlou05, line 88: There is an `acceptance_radius` input given, please use that one

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,6 +80,31 @@ def run(
...
+            index = 0
+            for i in range(0, len(landing_pad_locations)):
```
maxlou05, line 90: We prefer to use a for-each loop in python, like `for landing_pad in landing_pad_locations`

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,6 +80,31 @@ def run(
...
+                dist = (landing_pad_locations[i].location_x - self.waypoint.location_x)**2+(landing_pad_locations[i].location_y - self.waypoint.location_y)**2
+                if dist < dist_min:
+                    dist_min = dist
+                    index = i
+
+            closest_landing_pad_location = landing_pad_locations[i]
```
maxlou05, line 96: This was probably intended to be index `[index]` rather than `[i]`

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,6 +80,31 @@ def run(
...
+
+            command = commands.Command.create_set_relative_destination_command(closest_landing_pad_location.location_x - self.waypoint.location_x, closest_landing_pad_location.location_y - self.waypoint.location_y)
```
maxlou05, line 98: Although you should be at the waypoint right now, it is probably more accurate to use the drone's position rather than the waypoint's location.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,6 +80,31 @@ def run(
...
+
+            self.has_found_landing_pad = True
+
+        elif report.status == drone_status.DroneStatus.HALTED and self.has_found_landing_pad:
+
+            self.has_sent_landing_command = True
+
+        if self.has_sent_landing_command:
+            command = commands.Command.create_land_command()
```
maxlou05, line 107: There is probably no ned to separate these two conditions, they both mean the same thing (landing the drone), so they should be combined.

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +95,39 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
-        boxes_cpu = ...
+        boxes_cpu = boxes_xyxy.detach().cpu().numpy()
 
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
         # Hint: .shape gets the dimensions of the numpy array
         # for i in range(0, ...):
         #     # Create BoundingBox object and append to list
         #     result, box = ...
+        for i in range(0, boxes_cpu.shape[0]):
```
maxlou05, line 122: Again, it is preferred to use a for each loop in Python

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +95,39 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+            if result:
+                bounding_boxes.append(box)
```
maxlou05, line 125: In statistics, we normally would discard the entire datapoint (in this case the whole image) if part of it has issues (ie one of the bounding boxes are broken).
You do not need to change anything here, just something to think about.

maxlou05 (commented):
Reviewed

modules/bootcamp/decision_simple_waypoint.py
Cyuber: nevermind i rewrote a solution that always takes into account the global position

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -69,6 +76,16 @@ def run(
         # ============
 
         # Do something based on the report and the state of this class...
+        if report.status == drone_status.DroneStatus.HALTED:
+            # Defined here so that report object can be used
+            self.commands = [
+                commands.Command.create_set_relative_destination_command(self.waypoint.location_x - report.position.location_x, self.waypoint.location_y - report.position.location_y)
+            ]
```
maxlou05, line 83: This works, but since there's only 1 task (or a low finite number of tasks), you can just use a boolean variable rather than putting commands in a list.
(You do not need to change anything here, just something to think about)

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -38,6 +38,13 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
         # ============
 
         # Add your own
+        self.command_index = 0
+
+        self.commands = []
+
+        self.has_sent_landing_command = False
+
+        self.counter = 0
```
maxlou05, line 47: These variables are not used, so you can delete them

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +95,34 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Hint: .shape gets the dimensions of the numpy array
         # for i in range(0, ...):
         #     # Create BoundingBox object and append to list
         #     result, box = ...
-
-        return [], image_annotated
-        # ============
-        # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
-        # ============
+        for i in boxes_cpu.shape[0]:
```
maxlou05, line 122: `boxes_cpu.shape[0]` should be a number. So in this case you would have to use `range()`, as shown in the hint

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,6 +77,32 @@ def run(
...
+            # Defined here so that report object can be used
+            self.commands = [
+                commands.Command.create_set_relative_destination_command(self.waypoint.location_x, self.waypoint.location_y)
+            ]
+            if self.command_index < len(self.commands):
+                command = self.commands[self.command_index]
+                self.command_index += 1
+            elif not self.has_found_landing_pad:
+                # uses simple minimum value search to find closest pad and goes there
+                dist_min = 30000
```
maxlou05, line 90: Please use `float(infinity)`, what if the distance is greater than this?

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -38,6 +38,14 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
...
 
         # Add your own
+        self.command_index = 0
+
+        self.commands = []
+
+        self.has_sent_landing_command = False
+        self.has_found_landing_pad = False
+
+        self.counter = 0
```
maxlou05, line 48: This variable is not used, you can delete it

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,6 +77,32 @@ def run(
...
+                commands.Command.create_set_relative_destination_command(self.waypoint.location_x, self.waypoint.location_y)
+            ]
+            if self.command_index < len(self.commands):
+                command = self.commands[self.command_index]
+                self.command_index += 1
+            elif not self.has_found_landing_pad:
+                # uses simple minimum value search to find closest pad and goes there
+                dist_min = 30000
+                index = 0
+                for i in len(landing_pad_locations):
```
maxlou05, line 92: This for loop can be written as a for-each loop `for landing_pad in landing_pad_locations:`

maxlou05 (commented):
Reviewed

maxlou05 (commented):
Approved

PR #84, opened 2024-09-09
Everything Completed 🤩 

Zenkqi:
Add your description here...

### DIFF
--- a/modules/bootcamp/decision_simple_waypoint.py
+++ b/modules/bootcamp/decision_simple_waypoint.py
@@ -45,8 +45,6 @@ class DecisionSimpleWaypoint(base_decision.BaseDecision):
             self.waypoint.location_x, self.waypoint.location_y
         )
 
-        self.starting_halt = True
-
         # ============
         # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
         # ============
@@ -76,13 +74,19 @@ class DecisionSimpleWaypoint(base_decision.BaseDecision):
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
+        dist_to_waypoint_x = self.waypoint.location_x - report.position.location_x
+        dist_to_waypoint_y = self.waypoint.location_y - report.position.location_y
+
         # Do something based on the report and the state of this class...
-        if report.status == drone_status.DroneStatus.HALTED and self.starting_halt:
-            command = self.commands
-            self.starting_halt = False
-        elif report.status == drone_status.DroneStatus.HALTED and not self.has_sent_landing_command:
-            command = commands.Command.create_land_command()
+        if report.status == drone_status.DroneStatus.HALTED and not self.has_sent_landing_command:
             self.has_sent_landing_command = True
+            command = self.commands
+
+        elif (
+            dist_to_waypoint_x**2 + dist_to_waypoint_y**2 < self.acceptance_radius**2
+            and self.has_sent_landing_command
+        ):
+            command = commands.Command.create_land_command()
 
         # ============
         # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
--- a/modules/bootcamp/decision_simple_waypoint.py
+++ b/modules/bootcamp/decision_simple_waypoint.py
@@ -39,6 +39,12 @@ class DecisionSimpleWaypoint(base_decision.BaseDecision):
 
         # Add your own
 
+        self.has_sent_landing_command = False
+
+        self.commands = commands.Command.create_set_relative_destination_command(self.waypoint.location_x, self.waypoint.location_y)
+
+        self.starting_halt = True
+
         # ============
         # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
         # ============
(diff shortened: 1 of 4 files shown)


modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -69,6 +77,12 @@ def run(
         # ============
 
         # Do something based on the report and the state of this class...
+        if report.status == drone_status.DroneStatus.HALTED and self.starting_halt:
+            command = self.commands
+            self.starting_halt = False
+        elif report.status == drone_status.DroneStatus.HALTED and not self.has_sent_landing_command:
+            command = commands.Command.create_land_command()
+            self.has_sent_landing_command = True
```
AaronWang04, line 85: what is the purpose of self.starting_halt and self.has_sent_landing_command? how does it help the drone fly to the waypoint.

from what I understand, the drone halts, then u send.a null command followed by a land comman. but what is the purpose of the first null command. also, there is a self.accepted_radius member variable. implement using it but also think about why its there

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -38,6 +38,10 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
         # ============
 
         # Add your own
+        self.travel_to = commands.Command.create_set_relative_destination_command
```
AaronWang04, line 41: creating a wrapper for a function just creates more confusion, you should just call the function. if you want to make it clear that its "travel_to", then use comments

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +98,39 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
+
         # Hint: .shape gets the dimensions of the numpy array
         # for i in range(0, ...):
         #     # Create BoundingBox object and append to list
         #     result, box = ...
 
-        return [], image_annotated
+        for i in range(0, boxes_cpu.shape[0]):
```
AaronWang04, line 128: 0 is redundant

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +98,39 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+            if result:
+                bounding_boxes.append(box)
```
AaronWang04, line 131: alternatively, you can throw an error here, that is how its commonly handled in statistics. but that is a design choice and smoething to think about

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -67,8 +71,31 @@ def run(
...
+
+                    if curr_norm < min_norm:
+                        min_norm, min_location = curr_norm, (x, y)
+
+                command = self.travel_to(min_location[0], min_location[1])
+                self.halt_counter += 1
+
+            elif self.halt_counter == 2 and not self.has_sent_landing_command:
+                command = commands.Command.create_land_command()
+                self.has_sent_landing_command = True
```
AaronWang04, line 98: decent control flow. in this case halt counter makes sense to me. alternatiely, I think its better convention to have two separaete booleans with better naming in this case but thats up to u. it would be difficult for another person to navigate this code without comments (ideally, the best code are the ones where you don't have to write comments for)

AaronWang04: but again, like task 1, you have to use accepted_radius rather than halted to check for whether drone is at correct location

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -39,6 +39,12 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
 
         # Add your own
 
+        self.has_sent_landing_command = False
+
+        self.commands = commands.Command.create_set_relative_destination_command(
```
AaronWang04, line 44: this variable is unnecessary in this context (small script). in the event you want to use it, you should choose better variable names

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -38,6 +38,14 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
         # ============
 
         # Add your own
+        self.commands = commands.Command.create_set_relative_destination_command(
```
AaronWang04, line 41: same comment as above

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -38,6 +38,14 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
...
+            self.waypoint.location_x, self.waypoint.location_y
+        )
+
+        self.has_sent_landing_command = False
```
AaronWang04, line 45: this variable name is very clear on what it conveys

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -38,6 +38,14 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
...
+
+        self.min_location = [float("inf"), float("inf")]
+        self.starting_halt, self.waypoint_halt = True, True
```
AaronWang04, line 48: these are worse as they are more ambiguous on what the boolean value conveys, a good example could be reached_initial_halt

modules/bootcamp/decision_waypoint_landing_pads.py
AaronWang04, line 1: overall working from what I can tell, just work on variable naming. also, this can be implemented with a simpler control loop, you should think about how you can make this simpler but its not necessary for you to implement it for this bootcamp

AaronWang04 (approved):

PR #94, opened 2024-09-13
Finished... Probably not the cleanest but eh

Person8790:
Blake8 why no work

### DIFF
--- a/modules/bootcamp/decision_simple_waypoint.py
+++ b/modules/bootcamp/decision_simple_waypoint.py
@@ -32,13 +32,28 @@ class DecisionSimpleWaypoint(base_decision.BaseDecision):
         print(f"Waypoint: {waypoint}")
 
         self.acceptance_radius = acceptance_radius
-
         # ============
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
-        # Add your own
+    def relative_location_from_waypoint(self, current_x, current_y):
+        """
+        Finds how far you are from the waypoint based off your current location
+        """
+        waypoint_x, waypoint_y = self.waypoint.location_x, self.waypoint.location_y
 
+        return (waypoint_x - current_x), (waypoint_y - current_y)
+
+    def at_waypoint(self, current_x, current_y):
+        """
+        Returns if you are at the waypoint or not
+        """
+        dist_to_waypoint = sum(
+            dist**2 for dist in self.relative_location_from_waypoint(current_x, current_y)
+        )
+        if dist_to_waypoint**0.5 <= self.acceptance_radius:
+            return True
+        return False
         # ============
         # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
         # ============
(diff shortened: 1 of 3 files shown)


modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -32,13 +32,28 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
...
+        return (waypoint_x - current_x), (waypoint_y - current_y)
 
+    def at_waypoint(self, current_x: float, current_y: float) -> bool:
+        """
+        Returns if you are at the waypoint or not
+        """
+        dist_to_waypoint = sum(
+            dist**2 for dist in self.relative_location_from_waypoint(current_x, current_y)
+        )
+        if dist_to_waypoint**0.5 <= self.acceptance_radius:
```
maxlou05, line 54: Rather than taking a square root on the "left side", can you think of doing something else which is computationally easier?

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -35,9 +35,59 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
...
+
+        self.landing_pad = None
+        self.destination = self.waypoint
+        self.arrived_at_waypoint = False
+
         # ============
 
-        # Add your own
+    def find_closest_landing_pad(
+        self, landing_pad_list: location.Location, current_x: float, current_y: float
```
maxlou05, line 46: Great use of type hinting, but I think `landing_pad_list` is meant to be a list of Locations, and not just 1 location

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +118,33 @@ def run(
...
+            )
+            command = commands.Command.create_set_relative_destination_command(
+                relative_x_dist_to_destination, relative_y_dist_to_destination
+            )
+
+        elif report.status == drone_status.DroneStatus.HALTED and self.at_location(
+            current_x, current_y, self.destination
+        ):
+
+            if self.arrived_at_waypoint:
```
maxlou05, line 138: Although this works, it is a little confusing to readers because it looks like you are lading at the waypoint and not the landing pad. It would be better to rename this variable to something else like travelling_to_landing_pad.

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +94,38 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
+
         # Hint: .shape gets the dimensions of the numpy array
         # for i in range(0, ...):
         #     # Create BoundingBox object and append to list
         #     result, box = ...
+        for i in range(boxes_cpu.shape[0]):
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+            if result:
```
maxlou05, line 125: In statistics, we usually discard the entire data point (the whole image) if part of it is broken (the bounding box), and not just that one bounding box.

maxlou05 (commented):
Reviewed

maxlou05 (commented):
Approved!

PR #97, opened 2024-09-14
Daniel Suzuki's Bootcamp

1Two3Four:


modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -37,7 +37,14 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
-        # Add your own
+        self.command_index = 0
+        self.commands = [
+            commands.Command.create_set_relative_destination_command(
+                waypoint.location_x, waypoint.location_y
```
maxlou05, line 43: This command uses relative distance, so your position matters. You may end up going somewhere you didn't intend to.

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -69,6 +76,22 @@ def run(
...
+        if report.status == drone_status.DroneStatus.HALTED and self.command_index < len(
+            self.commands
+        ):
+            command = self.commands[self.command_index]
+            self.command_index += 1
+        elif (
+            report.status == drone_status.DroneStatus.HALTED
+            and (
+                (report.position.location_x - self.waypoint.location_x) ** 2
+                + (report.position.location_y - self.waypoint.location_y)
```
maxlou05, line 88: I think you forgot to square the delta y

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -69,6 +76,22 @@ def run(
...
+            )
+            <= self.acceptance_radius**2
+        ):
+            command = commands.Command.create_land_command()
+
+            self.has_sent_landing_command = True
```
maxlou05, line 94: This variable isn't really used for anything, so you can delete it

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -67,8 +70,55 @@ def run(
         # ============
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
+        def square(number: int) -> float:
```
maxlou05, line 73: number probably should be a float

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -67,8 +70,55 @@ def run(
...
+            return number**2
+
+        def squared_distance_from_position(
+            point: list[location.Location], current_position: list[location.Location]
```
maxlou05, line 77: point and position should probably just be 1 location, and not a list of locations

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -67,8 +70,55 @@ def run(
...
+            )
+            return distance
 
         # Do something based on the report and the state of this class...
+        if report.status == drone_status.DroneStatus.HALTED:
+
+            if self.halt_at_initialization:
+
+                command = commands.Command.create_set_relative_destination_command(
+                    self.waypoint.location_x, self.waypoint.location_y
```
maxlou05, line 90: This is a relative distance comamnd, please take into consideration your own position

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -67,8 +70,55 @@ def run(
...
+
+                    if (
+                        squared_distance_from_position(landing_pad, self.waypoint)
+                        < self.shortest_distance[0] ** 2 + self.shortest_distance[1] ** 2
+                    ):
+
+                        self.shortest_distance[0], self.shortest_distance[1] = (
+                            landing_pad.location_x - report.position.location_x,
+                            landing_pad.location_y - report.position.location_y,
+                        )
```
maxlou05, line 106: You can probably just store the distance squared instead of both dx and dy and calculating the distance again many times

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -67,8 +70,55 @@ def run(
...
+                        self.closest_pad = landing_pad
+
+                command = commands.Command.create_set_relative_destination_command(
+                    self.shortest_distance[0], self.shortest_distance[1]
```
maxlou05, line 110: Because here, you can just use self.closest_pad, which is a Location (has its own x and y coords)

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -67,8 +70,55 @@ def run(
...
+                )
+                self.is_halt_at_waypoint = False
+
+            elif (report.position.location_x**2 + report.position.location_y**2) / (
+                self.closest_pad.location_x**2 + self.closest_pad.location_y**2
+            ) <= (
+                (1 + self.acceptance_radius) ** 2
+            ):  # checks if current position is in acceptable radius of landing pad by making them a ratio
```
maxlou05, line 118: Not sure why you calculated like this ratio? You can just use the same way you did earlier with the comparing your distance to acceptance_radius

maxlou05 (commented):
Reviewed

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -37,7 +37,8 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
-        # Add your own
+        self.halt_at_initialization = True
+        self.has_sent_landing_command = False
```
maxlou05, line 41: You can delete this here too

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -67,8 +71,52 @@ def run(
...
+                command = commands.Command.create_set_relative_destination_command(
+                    self.closest_pad.location_x - report.position.location_x,
+                    self.closest_pad.location_y - report.position.location_y,
+                )
+                self.is_halt_at_waypoint = False
+
+            elif (
+                (report.position.location_x - self.closest_pad.location_x) ** 2
+                + (report.position.location_y - self.closest_pad.location_y) ** 2
+            ) <= ((self.acceptance_radius) ** 2):
```
maxlou05, line 117: I think your brackets are wrong here, your elif ends before the <= operator

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -67,8 +71,52 @@ def run(
...
+                self.is_halt_at_waypoint = True
+
+            elif self.is_halt_at_waypoint:
+                for landing_pad in landing_pad_locations:
+                    distance_from_position = squared_distance_from_position(
+                        landing_pad, self.waypoint
+                    )
+                    current_shortest_distance = (
+                        self.closest_pad.location_x - report.position.location_x
+                    ) ** 2 + (self.closest_pad.location_y - report.position.location_y) ** 2
```
maxlou05, line 104: You can use your square_distance_... function here too

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -36,12 +36,16 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
         # ============
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
+        # set as closest landing to waypoint
+        self.has_sent_landing_command = False
```
maxlou05, line 40: Probably don't need this one either

maxlou05 (commented):
Reviewed

modules/bootcamp/decision_waypoint_landing_pads.py
1Two3Four: I've realized that when i use black modules , the brackets that I add at the front of 'elif' and after '**2' end are removed. Is there any way to resolve this?

maxlou05 (commented):
Approved!

PR #98, opened 2024-09-15
Adrian's Bootcamp

agslaj-ALock:


modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -4,6 +4,8 @@
 Travel to designated waypoint.
 """
 
+import math
```
maxlou05, line 7: For this bootcamp, you are not allowed to import any libraries. It is possible to complete the bootcamp without using the math library. Think of how you can compare distances without using the square root function

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -4,6 +4,8 @@
 Travel to designated waypoint and then land at a nearby landing pad.
 """
 
+import math
```
maxlou05, line 7: Same, don't need to import library

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,9 +72,57 @@ def run(
         # ============
 
         # Do something based on the report and the state of this class...
+        status = report.status
+
+        if self.at_waypoint:
+            self.landing_pad = self.get_nearest_landing(report.position, landing_pad_locations)
```
maxlou05, line 78: This will calculate the nearest landing pad constantly while you are moving towards it, which is not necessary since you will never get closer to a different pad while travelling in a straight line towards the landing pad you selected. You can move this line a different section to only calculate it once (Think what step comes right before this one)

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +98,37 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # move the copy into CPU space,
         # and convert to a numpy array
-        boxes_cpu = ...
+        boxes_cpu = boxes_xyxy.detach().cpu().numpy()
 
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
+        for i in range(0, boxes_cpu.shape[0]):
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+            if result:
```
maxlou05, line 124: In statistics, we normally discard the entire datapoint (the whole image) if part of it goes wrong (the bounding box).

maxlou05 (commented):
Reviewed

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +98,37 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+        )
 
         # Get the Result object
-        prediction = ...
+        prediction = predictions[0]
 
         # Plot the annotated image from the Result object
         # Include the confidence value
-        image_annotated = ...
+        image_annotated = prediction.plot()
```
maxlou05, line 110: Please incldue the confidence value with the kwarg conf. You can also look at the ultralytics docs

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,9 +70,66 @@ def run(
         # ============
 
         # Do something based on the report and the state of this class...
+        status = report.status
+
+        if self.at_waypoint:
+            if status == drone_status.DroneStatus.HALTED:
+                self.landing_pad = self.get_nearest_landing(report.position, landing_pad_locations)
```
maxlou05, line 77: By putting it right after setting self.at_wayptoint to true, you only calculate this once rather than all the time continuously, which is unecessary.

maxlou05 (commented):
Reviewed

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +98,37 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+        )
 
         # Get the Result object
-        prediction = ...
+        prediction = predictions[0]
 
         # Plot the annotated image from the Result object
         # Include the confidence value
-        image_annotated = ...
+        image_annotated = prediction.plot(conf=0.75)
```
maxlou05, line 110: conf should be a boolean, not a float. It can either be included in the picture or not. Just set it to true

maxlou05 (commented):
Reviewed

maxlou05 (commented):
Approved

PR #102, opened 2024-09-19
Lucas Kim's bootcamp

Raptors65:


modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,7 +66,22 @@ def run(
...
-        # Do something based on the report and the state of this class...
+        allowed_error = 0.1
+
+        if report.status == drone_status.DroneStatus.HALTED:
+            # Calculate relative x and y distance required to reach waypoint
+            relative_x = self.waypoint.location_x - report.position.location_x
+            relative_y = self.waypoint.location_y - report.position.location_y
+
+            # Check if the waypoint has already been reached
+            if abs(relative_x) < allowed_error and abs(relative_y) < allowed_error:
```
maxlou05, line 77: Normally, we use Euclidean distance (L2 norm) rather than taxicab/manhattan distance (L1 norm). Try to use that instead

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -4,6 +4,8 @@
 Travel to designated waypoint and then land at a nearby landing pad.
 """
 
+import math
```
maxlou05, line 7: In this bootcamp, you are not allowed to import any libraries. The bootcamp can be completed without importing math or using square roots. You can think about what is your purpose for calculating distance

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +97,38 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         bounding_boxes = []
         # Hint: .shape gets the dimensions of the numpy array
-        # for i in range(0, ...):
-        #     # Create BoundingBox object and append to list
-        #     result, box = ...
+        for xyxy_box in boxes_cpu:
+            # Create BoundingBox object and append to list
+            _, box = bounding_box.BoundingBox.create(xyxy_box)
+
+            if box is None:
```
maxlou05, line 126: We have a "success" indicator returned from the create function, you can use that instead. It's more readable and can handle more cases than box does not exist. Maybe the box exists but is malformed.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +70,48 @@ def run(
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
-        # Do something based on the report and the state of this class...
+        if report.status == drone_status.DroneStatus.HALTED:
+            # Check if the waypoint has already been reached
+            if self.started_moving_to_landing_pad:
```
maxlou05, line 75: Although in this bootcamp, you are guarantee to make it to your destination, sometimes due to wind you end up somewhere else. Please use the acceptance_radius just like you used the "error" from the other task.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +70,48 @@ def run(
...
+                command = commands.Command.create_land_command()
+            elif self.started_moving_to_waypoint:
```
maxlou05, line 77: Same here, please confirm that you are indeed within the acceptance radius of the waypoint

maxlou05 (commented):
Reviewed

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +67,52 @@ def run(
...
+
+        if report.status == drone_status.DroneStatus.HALTED:
+            closest_landing_pad = min(
+                landing_pad_locations,
+                key=lambda pad: self.__squared_distance(report.position, pad),
+            )
+
+            # Check if the landing pad or waypoint has already been reached
+            if (
+                self.__squared_distance(report.position, closest_landing_pad) < allowed_error
```
maxlou05, line 80: For this, the function `run()` has a parameter `acceptance_radius` which you should use. The "error" should thus be squared for this case.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +67,52 @@ def run(
...
+                and self.has_reached_waypoint
+            ):
+                command = commands.Command.create_land_command()
+            elif self.__squared_distance(report.position, self.waypoint) < allowed_error:
```
maxlou05, line 84: Same here

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +67,52 @@ def run(
...
         # ============
 
-        # Do something based on the report and the state of this class...
+        allowed_error = 0.01
+
+        if report.status == drone_status.DroneStatus.HALTED:
+            closest_landing_pad = min(
+                landing_pad_locations,
+                key=lambda pad: self.__squared_distance(report.position, pad),
+            )
```
maxlou05, line 76: I think you moved it outside of the conditions. It should probably be after line 84, when you have reached the waypoint.

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +97,37 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         bounding_boxes = []
-        # Hint: .shape gets the dimensions of the numpy array
-        # for i in range(0, ...):
-        #     # Create BoundingBox object and append to list
-        #     result, box = ...
+        for xyxy_box in boxes_cpu:
+            # Create BoundingBox object and append to list
+            success, box = bounding_box.BoundingBox.create(xyxy_box)
+
+            if not success:
```
maxlou05, line 125: In statistics, we normally discard the entire data point (the whole image) if part of it has gone wrong (a bounding box). 

maxlou05 (commented):
Reviewed

maxlou05 (commented):
Approved

PR #103, opened 2024-09-20
Lucas Liao's Bootcamp

lucasliao0403:


modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -36,9 +36,19 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
         # ============
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
-
         # Add your own
 
+        self.command_index = 0
+        self.commands = [
+            commands.Command.create_set_relative_destination_command(
+                waypoint.location_x, waypoint.location_y
```
maxlou05, line 44: This is a relative destination command, so your own position matters. You may end up going somewhere you didn't expect.

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -67,9 +77,28 @@ def run(
...
+
+        if report.status == drone_status.DroneStatus.HALTED and self.command_index < len(
+            self.commands
+        ):
+
+            print(f"Halted at: {report.position}")
+
+            command = self.commands[self.command_index]
+            self.command_index += 1
+        elif report.status == drone_status.DroneStatus.HALTED and not self.has_sent_landing_command:
```
maxlou05, line 93: Although this can work for the bootcamp, what happens if the drone halts somewhere while it was on its way? Try to find a way to make sure it gets there.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -39,6 +39,19 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
 
         # Add your own
 
+        self.waypoint_reached = False
+
+        self.command_index = 0
+        self.commands = [
+            commands.Command.create_set_relative_destination_command(
+                waypoint.location_x, waypoint.location_y
```
maxlou05, line 47: Same, relative command

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,7 +81,53 @@ def run(
...
+        if (
+            report.status == drone_status.DroneStatus.HALTED
+            and not self.waypoint_reached
+            and self.command_index < len(self.commands)
+        ):
+            print(f"Halted at: {report.position}")
+            command = self.commands[self.command_index]
+            self.command_index += 1
+
+        elif report.status == drone_status.DroneStatus.HALTED and not self.waypoint_reached:
```
maxlou05, line 95: Again, try to find a different way to guarantee the drone gets to the waypoint

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,7 +81,53 @@ def run(
...
+            print(landing_pad_locations)
+            print("TARGET INDEX:", index)
+
+            self.commands.append(
+                commands.Command.create_set_relative_destination_command(
+                    landing_pad_locations[index].location_x - report.position.location_x,
+                    landing_pad_locations[index].location_y - report.position.location_y,
+                )
+            )
+            command = self.commands[self.command_index]
```
maxlou05, line 118: You could just set command to your desired command rather than adding it to a list and then reading it from there.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,7 +81,53 @@ def run(
...
+
+        elif (
```
maxlou05, line 120: Again, please ensure drone has reached the landing pad

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +96,37 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
         # * conf
         # * device
         # * verbose
-        predictions = ...
+        # predictions = ...
 
         # Get the Result object
-        prediction = ...
+        prediction = self.__model.predict(source=image)
```
maxlou05, line 102: This was supposed to be the `predictions` variable, and then take `prediction` to be the 0th index. Also, please read the comments above for which kwargs to include

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +96,37 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+
+        # print('\n\nHEYYYYYYYYYYY\n\n')
+        # print(prediction[0])
 
         # Plot the annotated image from the Result object
         # Include the confidence value
-        image_annotated = ...
+        image_annotated = prediction[0].orig_img
```
maxlou05, line 109: Please use the plot function, and include the confidence value, not just the image

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +96,37 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Get the xyxy boxes list from the Boxes object in the Result object
-        boxes_xyxy = ...
+        boxes_xyxy = prediction[0].boxes
 
         # Detach the xyxy boxes to make a copy,
         # move the copy into CPU space,
         # and convert to a numpy array
-        boxes_cpu = ...
+
+        boxes_cpu = boxes_xyxy.xyxy.cpu().numpy()
```
maxlou05, line 118: You can put the xyxy in the previous variable. Here, you should also detach to make a copy before moving it to CPU

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +96,37 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         bounding_boxes = []
         # Hint: .shape gets the dimensions of the numpy array
-        # for i in range(0, ...):
-        #     # Create BoundingBox object and append to list
-        #     result, box = ...
-
-        return [], image_annotated
+        for i in range(np.shape(boxes_cpu)[0]):
+            # Create BoundingBox object and append to list
+            temp_box = bounding_box.BoundingBox.create(np.array(boxes_cpu[i]))[1]
```
maxlou05, line 126: Please use the result variable rather than discarding it. It contains important information if the creation succeeded or not. You also don't need to convert a numpy array to a numpy array, you already did so in the last step.

maxlou05 (commented):
Reviewed

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -63,12 +75,71 @@ def run(
...
+                    index = i
+
+            print(landing_pad_locations)
+            print("TARGET INDEX:", index)
+
+            command = commands.Command.create_set_relative_destination_command(
+                landing_pad_locations[index].location_x - report.position.location_x,
+                landing_pad_locations[index].location_y - report.position.location_y,
+            )
+            # print("HALTED AND WAYPOINT REACHED\n\n\n\n\n\n\n\n")
```
maxlou05, line 124: You do not need to keep commented code, since git will save history. You can just delete it. You also seem to have a lot of print statements that print every iteration of run, which will be a lot of messages really fast. You can probably delete those too if you don't need it, as it won't clutter your screen. (also production doesn't need debug prints)

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +96,41 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
-        predictions = ...
+        predictions = self.__model.predict(
+            source=image, conf=0.7, device=self.__DEVICE, verbose=True
+        )
 
         # Get the Result object
-        prediction = ...
+        prediction = predictions[0]
+
+        # print('\n\nHEYYYYYYYYYYY\n\n')
```
maxlou05, line 106: Same, can just delete these commented print statements

maxlou05 (commented):
reviewed

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -36,13 +36,24 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
...
+        self.counter = 0
+
         # ============
         # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
         # ============
 
+    @staticmethod
+    def is_close(x: float, y: float, target_x: float, target_y: float, tolerance: float) -> bool:
+        """Determines if coordinate x,y, is within (euclidean distance) tolerence of coordinate target_x, target_y."""
+        dist = pow(pow((x - target_x), 2) + pow((y - target_y), 2), 0.5)
```
maxlou05, line 54: We would like to not use the square root function because it is costly. Can you determine if something is closer than the other without square rooting?

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -39,10 +39,26 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
...
+        self.index = -1
+
         # ============
         # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
         # ============
 
+    @staticmethod
+    def is_close(x: float, y: float, target_x: float, target_y: float, tolerance: float) -> bool:
+        """Determines if coordinate x,y, is within (euclidean distance) tolerence of coordinate target_x, target_y."""
+        dist = pow(pow((x - target_x), 2) + pow((y - target_y), 2), 0.5)
```
maxlou05, line 59: Same here, no need to square root

maxlou05 (commented):
reviewed

maxlou05:
Approved

PR #104, opened 2024-09-24
Sohams Bootcamp

SohamD1:


modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +98,39 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
-        boxes_cpu = ...
+        boxes_cpu = boxes_xyxy.detach().cpu().numpy()
 
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
         # Hint: .shape gets the dimensions of the numpy array
         # for i in range(0, ...):
         #     # Create BoundingBox object and append to list
         #     result, box = ...
+        for boxes_xyxy in boxes_cpu:
```
maxlou05, line 127: I would suggest you not to use the same variable name here as the one you had before. It can be confusing for readers, and maybe even cause problems later on

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +98,39 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+            success, box = bounding_box.BoundingBox.create(boxes_xyxy)
+            if not success:
+                continue
 
         return [], image_annotated
```
maxlou05, line 132: You should return the bounding_boxes array that you just filled, or else you don't detect any landing pads

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -1,77 +1,143 @@
...
-        # ============
+        # Group related attributes into a single dictionary to reduce attribute count
+        self.state = {"waypoint_found": False, "landing_pad_found": False, "closest_pad": None}
+
+        self.has_sent_landing_command = False
+        self.reached_waypoint = False
+        self.moving_to_landing_pad = False
+        self.counter = 0
+
+    def at_point(self, current_x: float, current_y: float) -> bool:
```
maxlou05, line 40: You never ended up using this function. You can probably remove it (since it's the same as the is_within_acceptance_radius function)

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +98,39 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
         # Hint: .shape gets the dimensions of the numpy array
         # for i in range(0, ...):
         #     # Create BoundingBox object and append to list
         #     result, box = ...
+        for boxes_xyxy in boxes_cpu:
+            success, box = bounding_box.BoundingBox.create(boxes_xyxy)
+            if not success:
+                continue
```
maxlou05, line 130: Normally in statistics, we discard the entire data point (the entire image) if part of it goes wrong (a bounding box).

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -1,77 +1,143 @@
...
-        # ============
-        # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
-        # ============
+        # Group related attributes into a single dictionary to reduce attribute count
+        self.state = {"waypoint_found": False, "landing_pad_found": False, "closest_pad": None}
+
+        self.has_sent_landing_command = False
+        self.reached_waypoint = False
+        self.moving_to_landing_pad = False
+        self.counter = 0
```
maxlou05, line 38: I don't think you used any of these, since you used your state dictionary instead

maxlou05 (commented):
Reviewed

maxlou05 (commented):
reviewed

maxlou05 (commented):
Approved

PR #109, opened 2024-09-28
Bootcamp - Davis Liu

davisliu2006:


modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,6 +79,34 @@ def run(
         # ============
 
         # Do something based on the report and the state of this class...
+        pos_x = report.position.location_x
+        pos_y = report.position.location_y
+        if report.status == drone_status.DroneStatus.HALTED:  # if command done
```
AaronWang04, line 84: apply the double check with location here as well

modules/bootcamp/detect_landing_pad.py
```diff
@@ -91,38 +88,45 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
         # Hint: .shape gets the dimensions of the numpy array
         # for i in range(0, ...):
         #     # Create BoundingBox object and append to list
         #     result, box = ...
+        for box_cpu in boxes_cpu:
+            result, box = bounding_box.BoundingBox.create(box_cpu)
+            if not result:
+                return [], image_annotated  # data is invalid
```
AaronWang04, line 126: nice job returning the whole dataset as wrong, this is common practice done

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,7 +81,47 @@ def run(
...
+            elif self.command_index < len(self.commands):  # next command
+                print(f"cmd({self.command_index}) pos({report.position})")
+                command = self.commands[self.command_index]
+                if (
+                    command.get_command_type()
+                    == commands.Command.CommandType.SET_RELATIVE_DESTINATION
+                ):  # update target position
+                    cmd_x, cmd_y = command.get_relative_destination()
+                    self.trg_x += cmd_x
+                    self.trg_y += cmd_y
```
AaronWang04, line 118: I think this is redundant, you can just use the waypoint's location_x and location_y to check whether you are at the location you want since you know that is the exact location you are trying to reach

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,7 +81,47 @@ def run(
...
+                self.command_index += 1
+            elif self.land_status == 0:  # start landing
+                command = commands.Command.create_land_command()
+                self.land_status = 2
```
AaronWang04, line 122: nit pick but generally you want the "double check" if statements to be at the bottom, which makes more sense when reading

AaronWang04 (commented):
good job adding the check, just some nit picks on PR quality and code flow issues that can be improved!

AaronWang04 (approved):
LGTM, remember that you won't have to cache waypoint location for our specific scenario because waypoint is constant and is passed in as an argument.

PR #111, opened 2024-09-28
Owen's Bootcamp

wang-owen:


modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +95,42 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+        prediction = predictions[0]
 
         # Plot the annotated image from the Result object
         # Include the confidence value
-        image_annotated = ...
+        image_annotated = prediction.plot(conf=True)
 
         # Get the xyxy boxes list from the Boxes object in the Result object
-        boxes_xyxy = ...
+        boxes_xyxy = prediction.boxes.xyxy if prediction.boxes else []
```
TongguangZhang, line 114: just return empty list and image_annotated here if there are no boxes - we prefer early return rather than going thru the rest of the logic which basically does nothing (but has potential to error)

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +95,42 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+        print("boxes_xyxy:", boxes_xyxy)
 
         # Detach the xyxy boxes to make a copy,
         # move the copy into CPU space,
         # and convert to a numpy array
-        boxes_cpu = ...
+        boxes_cpu = [box.cpu().numpy() for box in boxes_xyxy]
```
TongguangZhang, line 120: you can call this on the whole matrix (another reason why u don't want to pass the empty list thru since it doesn't have that method)

TongguangZhang, line 120: make sure to also catch detach(), this removes it from the computational graph, which turns the gradients off, gradients are used in training for back propagation, but since we're doing inference here, we don't wanna keep tracking gradients

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +95,42 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Hint: .shape gets the dimensions of the numpy array
-        # for i in range(0, ...):
-        #     # Create BoundingBox object and append to list
-        #     result, box = ...
+        for i in range(0, len(boxes_xyxy)):
+            # Create BoundingBox object and append to list
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+
+            if result and box:
+                bounding_boxes.append(box)
```
TongguangZhang, line 131: instead of skipping bad bounding boxes, we want to skip the whole image if any box creation fails, in data science, we don't wanna return partial data (in this case 1 image is 1 data point, and the boxes are parts of that data)

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -69,6 +69,16 @@ def run(
         # ============
 
         # Do something based on the report and the state of this class...
+        if report.status == drone_status.DroneStatus.HALTED:
```
TongguangZhang, line 72: what if the drone is moving and within the acceptance radius (maybe some bug causes the drone to move slightly past the waypoint)?

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -38,6 +38,30 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
         # ============
 
         # Add your own
+        self.closest_landing_pad = None
+
+    def get_distance_squared(self, loc1: location.Location, loc2: location.Location) -> float:
```
TongguangZhang, line 43: `loc_1` instead of `loc1` - also, generally we prefer full var 😉  names (not a big deal tho)

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -38,6 +38,30 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
...
+        """Returns the distance squared between two locations"""
+        return abs((loc2.location_x - loc1.location_x) + (loc2.location_y - loc1.location_y))
```
TongguangZhang, line 45: seems like this isn't the square distance (this is still a valid way to measure - taxi drivers sometimes do this, not super ideal for this application imo but u can use it, just rename ur function)

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -38,6 +38,30 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
...
+        closest_landing_pad = None
+        for loc in landing_pad_locations:
+            dist = self.get_distance_squared(pos, loc)
+            if dist < shortest_dist:
+                shortest_dist = dist
+                closest_landing_pad = loc
+        return closest_landing_pad
+
+    def within_range(
+        self, loc: location.Location, target: location.Location, radius: float
```
TongguangZhang, line 60: don't pass in  the `radius` param if u don't need it

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,6 +93,28 @@ def run(
         # ============
 
         # Do something based on the report and the state of this class...
+        if report.status == drone_status.DroneStatus.HALTED:
```
TongguangZhang, line 96: similar to above, what if drone is moving and within radius?

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,6 +93,28 @@ def run(
...
+            if self.within_range(report.position, self.waypoint, self.acceptance_radius):
+                # Drone has reached its waypoint, calculate nearest landing pad
+                self.closest_landing_pad = self.get_closest_landing_pad(
+                    report.position, landing_pad_locations
+                )
+                if self.closest_landing_pad:
```
TongguangZhang, line 102: since this isn't a bool, it's better to check if it exists with `if not None`

TongguangZhang (commented):
reviewed

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -39,6 +39,14 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
 
         # Add your own
 
+    def within_range(
+        self, loc: location.Location, target: location.Location, radius: float
```
TongguangZhang, line 43: why pass in the radius when you already have it in the instance (through self)?

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -38,6 +38,37 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
...
+        closest_landing_pad = None
+        for loc in landing_pad_locations:
+            dist = self.get_distance_squared(pos, loc)
+            if dist < shortest_dist:
+                shortest_dist = dist
+                closest_landing_pad = loc
+        return closest_landing_pad
+
+    def within_range(
+        self, loc: location.Location, target: location.Location, radius: float
```
TongguangZhang, line 66: same as above

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +95,46 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+        if not prediction.boxes:
+            return [], image_annotated
+        boxes_xyxy = prediction.boxes.xyxy
 
         # Detach the xyxy boxes to make a copy,
         # move the copy into CPU space,
         # and convert to a numpy array
-        boxes_cpu = ...
+        boxes_cpu = (
+            boxes_xyxy.detach().cpu().numpy() if isinstance(boxes_xyxy, torch.Tensor) else []
```
TongguangZhang, line 122: if you want to do this check, again do not set it to [], just return [], imo this check isn't necessary since it's guaranteed that if it exists, it's a tensor

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +95,46 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
-        #     # Create BoundingBox object and append to list
-        #     result, box = ...
+        for box_cpu in boxes_cpu:
+            # Create BoundingBox object and append to list
+            result, box = bounding_box.BoundingBox.create(box_cpu)
+
+            if result and box:
+                bounding_boxes.append(box)
+            else:
+                return [], image_annotated
```
TongguangZhang, line 135: do not check if box exists - that's what the result flag is for

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +95,46 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         bounding_boxes = []
         # Hint: .shape gets the dimensions of the numpy array
-        # for i in range(0, ...):
-        #     # Create BoundingBox object and append to list
-        #     result, box = ...
+        for box_cpu in boxes_cpu:
+            # Create BoundingBox object and append to list
+            result, box = bounding_box.BoundingBox.create(box_cpu)
+
+            if result and box:
```
TongguangZhang, line 132: small optimization is to just check `if not result` and returning instead of having 2 branches, not a big deal tho

TongguangZhang (commented):
reviewed

TongguangZhang (commented):
lgtm

PR #113, opened 2024-10-02
Bootcamp - Catherine Jin

cathjin:


modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +95,39 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+            boxes_xyxy = prediction.boxes.xyxy
+            # Detach the xyxy boxes to make a copy,
+            # move the copy into CPU space,
+            # and convert to a numpy array
+            boxes_cpu = boxes_xyxy.detach().cpu().numpy()
+            # Loop over the boxes list and create a list of bounding boxes
+            for i in enumerate(boxes_cpu):
+                index = i[0]
+                (x_min, y_min, x_max, y_max) = boxes_cpu[index]
+                bounding_box_obj = bounding_box.BoundingBox.create(
```
maxlou05, line 121: You can directly pass in the `boxes_cpu[index]` object, no need to remake a numpy array

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +95,39 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+                    np.array([x_min, y_min, x_max, y_max])
+                )[1]
+                bounding_boxes.append(bounding_box_obj)
```
maxlou05, line 124: Please check the BoundingBox class, it returns a tuple.

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -67,9 +67,26 @@ def run(
         # ============
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
-
+        # MOVING = 0  # Moving towards destination
+        # HALTED = 1  # Not moving
+        # LANDED = 2  # On ground
         # Do something based on the report and the state of this class...
 
+        if (
```
maxlou05, line 75: This is good, but what if the drone halted in the middle of its path?

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -19,6 +19,19 @@
...
+def distance_to_landing_pad(
+    landing_pad_x: location.Location,
+    landing_pad_y: location.Location,
+    position_x: location.Location,
+    position_y: location.Location,
+) -> int:
+    """
+    Finds the distance to a landing pad
+    """
+    distance = ((landing_pad_x - position_x) ** 2 + (landing_pad_y - position_y) ** 2) ** (1 / 2)
```
maxlou05, line 31: Can you think of how not to use the square root function? It is more computationally heavy, and we do not need to use it

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,9 +82,45 @@ def run(
...
+            command = commands.Command.create_set_relative_destination_command(
+                self.waypoint.location_x, self.waypoint.location_y
+            )
+        elif (
+            report.status == report.status.HALTED
+            and report.position.location_x == self.waypoint.location_x
+            and report.position.location_y == self.waypoint.location_y
+        ):
+            command = commands.Command.create_halt_command()
+            min_distance = 100000000000
```
maxlou05, line 99: Please use `float("inf")` instead of an arbitrary large number

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,9 +82,45 @@ def run(
...
+                        landing_pad.location_y,
+                        report.position.location_x,
+                        report.position.location_y,
+                    )
+                    destination = landing_pad
+            command = commands.Command.create_set_relative_destination_command(
+                destination.location_x - report.position.location_x,
+                destination.location_y - report.position.location_y,
+            )
+        elif report.status == report.status.HALTED and report.position == destination:
```
maxlou05, line 121: Again, this is good, but what happens if the drone stops on the way to the waypoint? Or maybe it stopped on the way to the landing pad?

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -70,8 +110,59 @@ def run(
 
         # Do something based on the report and the state of this class...
 
+        if (
+            report.status == report.status.HALTED
+            and 0 <= abs(report.position.location_x) < abs(self.waypoint.location_x)
```
maxlou05, line 115: Please use euclidean distance

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -70,8 +110,59 @@ def run(
...
+            and 0 <= abs(report.position.location_y) < abs(self.waypoint.location_y)
+        ):
+            command = commands.Command.create_set_relative_destination_command(
+                self.waypoint.location_x, self.waypoint.location_y
```
maxlou05, line 119: Relative distance

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -70,8 +110,59 @@ def run(
...
+            )
+        elif (
+            report.status == report.status.HALTED
+            and (
+                self.waypoint.location_x
+                <= report.position.location_x
+                < closest_landing_pad(report, landing_pad_locations).location_x
+                or self.waypoint.location_x
+                >= report.position.location_x
+                > closest_landing_pad(report, landing_pad_locations).location_x
```
maxlou05, line 129: Again, please calculate the distances between the points rather than comparing their coordinates. You can compare distance squared, if that's what you were worried about. (a < b) => (a^2 < b^2) for any positive real number

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -70,8 +110,59 @@ def run(
...
+
+        elif (
+            report.status == report.status.HALTED
+            and (
+                abs(
+                    report.position.location_x
+                    - closest_landing_pad(report, landing_pad_locations).location_x
+                )
+            )
+            < 0.1
```
maxlou05, line 155: There is a `self.accpetance_radius`, which specifies how close the drone should get to the landing pad / waypoint

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +95,34 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Detach the xyxy boxes to make a copy,
         # move the copy into CPU space,
         # and convert to a numpy array
-        boxes_cpu = ...
-
+        boxes_cpu = boxes_xyxy.detach().cpu().numpy()
         # Loop over the boxes list and create a list of bounding boxes
-        bounding_boxes = []
+        for _, boxes in enumerate(boxes_cpu):
+            box = bounding_box.BoundingBox.create(boxes)[1]
```
maxlou05, line 118: Please use tuple unpacking here, we don't want to throw away the information (boolean) if the create was successful or not. We want to handle the case in which the create failed

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -70,8 +110,59 @@ def run(
...
+        if (
+            report.status == report.status.HALTED
+            and 0 <= abs(report.position.location_x) < abs(self.waypoint.location_x)
+            and 0 <= abs(report.position.location_y) < abs(self.waypoint.location_y)
+        ):
+            command = commands.Command.create_set_relative_destination_command(
+                self.waypoint.location_x, self.waypoint.location_y
+            )
+        elif (
+            report.status == report.status.HALTED
```
maxlou05, line 122: I think you need some sort of way to tell that you have been to the waypoint. As soon as you leave the waypoint, you will try to go back to the waypoint because you are no longer within acceptance radius of the waypoint

maxlou05 (commented):
Reviewed

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,9 +96,34 @@ def run(
...
+            and self.waypoint.location_x == report.position.location_x
+            and self.arrived_waypoint is False
+        ):
+            self.arrived_waypoint = True
+            command = commands.Command.create_set_relative_destination_command(
+                closest_landing_pad(report, landing_pad_locations).location_x
+                - report.position.location_x,
+                closest_landing_pad(report, landing_pad_locations).location_y
+                - report.position.location_y,
+            )
```
maxlou05, line 120: You can just save closest landing pad to a temporary variable instead of calculating it twice and writing all of this

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,9 +96,34 @@ def run(
...
+                report.position, closest_landing_pad(report, landing_pad_locations)
+            )
+            <= self.acceptance_radius**2
+            and self.arrived_waypoint is True
+        ):
+            command = commands.Command.create_land_command()
+
+        elif (
+            report.status == report.status.HALTED
+            and self.waypoint.location_x == report.position.location_x
```
maxlou05, line 111: Please use your distance and acceptance radius to determine if you have arrived at waypoint rather than this

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,9 +96,34 @@ def run(
...
+            self.arrived_waypoint = True
+            command = commands.Command.create_set_relative_destination_command(
+                closest_landing_pad(report, landing_pad_locations).location_x
+                - report.position.location_x,
+                closest_landing_pad(report, landing_pad_locations).location_y
+                - report.position.location_y,
+            )
+        elif report.status == report.status.HALTED and self.arrived_waypoint is False:
+            command = commands.Command.create_set_relative_destination_command(
+                self.waypoint.location_x, self.waypoint.location_y
```
maxlou05, line 123: Relative distance command

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +95,36 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
         # * conf
         # * device
         # * verbose
-        predictions = ...
+        bounding_boxes = []
+        predictions = self.__model.predict(
+            source=image, show=False, conf=0.7, device="cpu", verbose=True
```
maxlou05, line 100: Rather than hard-coding device, see line 39: __DEVICE

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +95,36 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # move the copy into CPU space,
         # and convert to a numpy array
-        boxes_cpu = ...
-
+        boxes_cpu = boxes_xyxy.detach().cpu().numpy()
         # Loop over the boxes list and create a list of bounding boxes
-        bounding_boxes = []
+        for i in range(0, boxes_xyxy.shape[0]):
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+            if result is True:
```
maxlou05, line 119: You can also just leave it as if result, since result is a boolean

yolov8n.pt
maxlou05, line 1: You shouldn't need to download another model, the bootcamp contains one

maxlou05 (commented):
Reviewed

maxlou05 (commented):
Approved

PR #115, opened 2024-10-14
Completed Autonomy Bootcamp

Aayush1104:


modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,8 +70,20 @@ def run(
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
-        # Do something based on the report and the state of this class...
-
+        while self.pause is True:
```
TongguangZhang, line 73: it doesn't seem like you ever set self.pause to False so the rest of your code wouldn't run?

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,8 +70,20 @@ def run(
...
+            return command
+        if report.status == drone_status.DroneStatus.HALTED and self.waypoint_reached is False:
+            radius_away_squared = (report.position.location_x - self.waypoint.location_x) ** 2 + (
+                report.position.location_y - self.waypoint.location_y
+            ) ** 2
+            if radius_away_squared > self.acceptance_radius**2:
```
TongguangZhang, line 79: we usually define acceptance radius as < is acceptable, >= is unacceptable

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,8 +70,20 @@ def run(
...
+                command = commands.Command.create_set_relative_destination_command(
+                    self.waypoint.location_x - report.position.location_x,
+                    self.waypoint.location_y - report.position.location_y,
+                )
+            elif radius_away_squared <= self.acceptance_radius**2:
```
TongguangZhang, line 84: while it's good to be specific in your cases - if it's not >, it must be <=, so this can just be an else

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -67,8 +69,44 @@ def run(
         # ============
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
-
-        # Do something based on the report and the state of this class...
+        # define a functin to calculate distance between each item in an array and the current location
+        def closest_landing_pad(
```
TongguangZhang, line 73: it's not great to define functions within functions - since you only need to define this function once, defining it multiple times is inefficient

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -67,8 +69,44 @@ def run(
...
+            for landing_pad in landing_pad_locations:
+                distance_squared = (report.position.location_x - landing_pad.location_x) ** 2 + (
+                    report.position.location_y - landing_pad.location_y
+                ) ** 2
+                if distance_squared < nearest_distance_squared:
+                    nearest_distance_squared = distance_squared
+                    nearest_landing_pad = landing_pad
+            return nearest_landing_pad
+
+        while self.pause is True:
```
TongguangZhang, line 87: here, self.pauses is never True, so we don't need this loop since it never runs

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -67,8 +69,44 @@ def run(
...
+        if report.status == drone_status.DroneStatus.HALTED and self.waypoint_reached is False:
+            radius_away_squared = (report.position.location_x - self.waypoint.location_x) ** 2 + (
+                report.position.location_y - self.waypoint.location_y
+            ) ** 2
+            if radius_away_squared > self.acceptance_radius**2:
+                command = commands.Command.create_set_relative_destination_command(
+                    self.waypoint.location_x - report.position.location_x,
+                    self.waypoint.location_y - report.position.location_y,
+                )
+            elif radius_away_squared <= self.acceptance_radius**2:
```
TongguangZhang, line 98: same as above, not > implies <=

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -67,8 +69,44 @@ def run(
...
+                    report.position.location_y - landing_pad.location_y
+                ) ** 2
+                if distance_squared < nearest_distance_squared:
+                    nearest_distance_squared = distance_squared
+                    nearest_landing_pad = landing_pad
+            return nearest_landing_pad
+
+        while self.pause is True:
+            return command
+        if report.status == drone_status.DroneStatus.HALTED and self.waypoint_reached is False:
```
TongguangZhang, line 89: it's more pythonic to say `not self.waypoint_reached` instead of `self.waypoint_reached is False`

TongguangZhang, line 89: since both branches first check `report.status == drone_status.DroneStatus.HALTED`, you should have both nested inside a larger if block instead of evaluating it twice

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -67,8 +69,44 @@ def run(
...
+                report.position.location_y - self.waypoint.location_y
+            ) ** 2
+            if radius_away_squared > self.acceptance_radius**2:
+                command = commands.Command.create_set_relative_destination_command(
+                    self.waypoint.location_x - report.position.location_x,
+                    self.waypoint.location_y - report.position.location_y,
+                )
+            elif radius_away_squared <= self.acceptance_radius**2:
+                self.waypoint_reached = True
+        elif report.status == drone_status.DroneStatus.HALTED and self.waypoint_reached is True:
```
TongguangZhang, line 100: simply `self.waypoint_reached` instead of `self.waypoint_reached is True`

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +98,43 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
-        # for i in range(0, ...):
-        #     # Create BoundingBox object and append to list
-        #     result, box = ...
+        for i in range(boxes_numpy.shape[0]):
+            box = boxes_numpy[i]
+
+            # Create BoundingBox object (update based on your actual constructor)
+            success, result = bounding_box.BoundingBox.create(box)
+            if success:
+                bounding_boxes.append(result)
```
TongguangZhang, line 135: instead of skipping failed box creations, we wanna skip the whole image if one of the boxes fail
in data science, we delete an entire data point if part of it is corrupted, instead of returning partial data - in this case, the data point is 1 image, so if some of the bounding boxes failed to create, we should skip the image

TongguangZhang (commented):
reviewed

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,8 +69,18 @@ def run(
...
+            radius_away_squared = (report.position.location_x - self.waypoint.location_x) ** 2 + (
+                report.position.location_y - self.waypoint.location_y
+            ) ** 2
+            if radius_away_squared >= self.acceptance_radius**2:
+                command = commands.Command.create_set_relative_destination_command(
+                    self.waypoint.location_x - report.position.location_x,
+                    self.waypoint.location_y - report.position.location_y,
+                )
+            else:
+                self.waypoint_reached = True
```
TongguangZhang, line 82: do you really need to set this variable? it seems like the distance check already covers this

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -37,12 +37,30 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
...
 
-        # Add your own
+        self.drone_status = None
+        self.waypoint_reached = False
 
         # ============
         # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
         # ============
 
+    def closest_landing_pad(
```
TongguangZhang, line 47: does this need the self parameter? if not, consider making it a static method

TongguangZhang (commented):
reviewed

PR #117, opened 2024-10-18
finish

zhengbrayden:


modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,7 +86,47 @@ def run(
...
+        elif report.status == drone_status.DroneStatus.HALTED:
+
+            if self.waypoint_reached:
+                # We are at landing pad, so land
+                return commands.Command.create_land_command()
+
+            # Get to the way point
+            # check if it is within the appropriate zone
+
+            if calc_dist(report.position, self.waypoint) < self.acceptance_radius**2:
```
TongguangZhang, line 101: what if you reach the waypoint, start moving, and then stop again - do you want to return to the waypoint and look for landing pads again?

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,7 +86,47 @@ def run(
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
-        # Do something based on the report and the state of this class...
+        if report.status == drone_status.DroneStatus.MOVING:
+            if calc_dist(report.position, report.destination) < self.acceptance_radius**2:
+                command = commands.Command.create_halt_command()
+        elif report.status == drone_status.DroneStatus.HALTED:
+
+            if self.waypoint_reached:
```
TongguangZhang, line 94: what if this is called while the drone is still halted at the waypoint? or if the drone stops before getting to the landing pad? make sure that you check the drone is within the acceptance radius for the landing pad too before landing

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +95,34 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
         # * conf
         # * device
         # * verbose
-        predictions = ...
+
+        predictions = self.__model.predict(source=image, conf=0.25, device=None, verbose=False)
```
TongguangZhang, line 99: 0.25 might be a little too low

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +95,34 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         bounding_boxes = []
         # Hint: .shape gets the dimensions of the numpy array
-        # for i in range(0, ...):
-        #     # Create BoundingBox object and append to list
-        #     result, box = ...
+        for i in range(0, boxes_cpu.shape[0]):
+            # Create BoundingBox object and append to list
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+            if result:
+                bounding_boxes.append(box)
```
TongguangZhang, line 123: instead of skipping failed box creations, we wanna skip the whole image if one of the boxes fail
in data science, we delete an entire data point if part of it is corrupted, instead of returning partial data - in this case, the data point is 1 image, so if some of the bounding boxes failed to create, we should skip the image

TongguangZhang (commented):
reviewed

modules/bootcamp/decision_waypoint_landing_pads.py
zhengbrayden: I think we would want to continue on to the landing pad right?

modules/bootcamp/decision_waypoint_landing_pads.py
zhengbrayden: Hm well if that happened then I guess it would just land right away somewhere between the waypoint and the landing pad. But I tried to implement my code so that once it reaches the waypoint, it would move directly to the landing pad, so I assumed that it wouldn't stop until it was at the landing pad. But I have modified the code in case it does stop in between.

modules/bootcamp/detect_landing_pad.py
zhengbrayden: changed to 0.5

modules/bootcamp/detect_landing_pad.py
zhengbrayden: Ah that makes sense. I have changed it so that it returns an empty list and numpy array

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +95,36 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
-        #     result, box = ...
-
-        return [], image_annotated
+        for i in range(0, boxes_cpu.shape[0]):
+            # Create BoundingBox object and append to list
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+            if result:
+                bounding_boxes.append(box)
+            else:
+                return [], np.array([])
```
TongguangZhang, line 125: return [], image_annotated

TongguangZhang (commented):
reviewed

TongguangZhang (commented):
lgtm

PR #119, opened 2024-10-28
Bootcamp Submission - Honzik Schenk

honzikschenk:


modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,7 +68,33 @@ def run(
...
+
+        # If the drone is halted and not at the waypoint, send a command to move to the waypoint
+        # Also handles if the drone is not on the target after landing
+        if (
+            report.status == drone_status.DroneStatus.HALTED
+            and (
+                direction_coordinate_x * direction_coordinate_x
+                + direction_coordinate_y * direction_coordinate_y
+            )
+            >= self.acceptance_radius * self.acceptance_radius
```
TongguangZhang, line 82: use `** 2` to make it clear you're doing the square, it also makes your code cleaner

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,7 +68,33 @@ def run(
...
+        ) or (
+            report.status == drone_status.DroneStatus.LANDED
+            and (
+                direction_coordinate_x * direction_coordinate_x
+                + direction_coordinate_y * direction_coordinate_y
+            )
+            >= self.acceptance_radius * self.acceptance_radius
```
TongguangZhang, line 89: it seems like this statement is unnecessary - the whole if statement simplifies to be "if inside radius and (landed or halted)"

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,7 +68,33 @@ def run(
...
+        ):
+            command = commands.Command.create_set_relative_destination_command(
+                direction_coordinate_x, direction_coordinate_y
+            )
+        # If the drone is halted, at the waypoint, and not already landed, send a command to land
+        elif report.status == drone_status.DroneStatus.HALTED and not self.has_sent_landing_command:
+            command = commands.Command.create_land_command()
+            self.has_sent_landing_command = True
```
TongguangZhang, line 97: what if the drone starts landing, and then stops for some reason, since you would want to send it another landing command to resume landing, do you need to set the `has_sent_landing_command`? also, do you need to set this flag to achieve your desired functionality?

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -37,12 +37,42 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
...
 
         # ============
         # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
         # ============
 
+    def get_squared_distance(self, x1: float, y1: float, x2: float, y2: float) -> float:
+        """
+        Calculate the distance between two points and return its square
+        """
+        return (x1 - x2) * (x1 - x2) + (y1 - y2) * (y1 - y2)
```
TongguangZhang, line 52: use `** 2`

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -37,12 +37,42 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
...
+
+    def get_closest_landing_pad(
+        self, report: drone_report.DroneReport, landing_pad_locations: "list[location.Location]"
+    ) -> int:
+        """
+        Find the closest landing pad to the drone.
+        """
+        closest_landing_pad = 0
```
TongguangZhang, line 60: are you guaranteed that there's something at the 0th position of the list of landing pads?

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -37,12 +37,42 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
...
+                landing_pad_locations[i].location_x,
+                landing_pad_locations[i].location_y,
+            ) < self.get_squared_distance(
+                report.position.location_x,
+                report.position.location_y,
+                landing_pad_locations[closest_landing_pad].location_x,
+                landing_pad_locations[closest_landing_pad].location_y,
+            ):
+                closest_landing_pad = i
+        return closest_landing_pad
```
TongguangZhang, line 74: instead of storing the index of the landing pad, you should store the landing pad object itself

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,7 +98,53 @@ def run(
...
-        # Do something based on the report and the state of this class...
+        # If the drone is at its desired location, then you must have reached the waypoint
+        if (
+            self.get_squared_distance(
+                report.position.location_x,
+                report.position.location_y,
+                self.waypoint.location_x,
+                self.waypoint.location_y,
+            )
+            < self.acceptance_radius * self.acceptance_radius
```
TongguangZhang, line 109: use `** 2`

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,7 +98,53 @@ def run(
...
+                direction_coordinate_x, direction_coordinate_y
+            )
+        # If the drone is halted, at the waypoint, not already landed, send a command to land
+        elif (
+            report.status == drone_status.DroneStatus.HALTED
+            and self.has_gone_to_waypoint
+            and not self.has_sent_landing_command
+        ):
+            command = commands.Command.create_land_command()
+            self.has_sent_landing_command = True
```
TongguangZhang, line 147: again, do you need to force it to only send the landing command once?

modules/bootcamp/detect_landing_pad.py
```diff
@@ -7,7 +7,6 @@
 import pathlib
 
 import numpy as np
-import torch
```
TongguangZhang, line 10: don't delete the imports

modules/bootcamp/detect_landing_pad.py
```diff
@@ -39,7 +34,8 @@ class DetectLandingPad:
     # Chooses the GPU if it exists, otherwise runs on the CPU
     # If you have a CUDA capable GPU but want to force it to
     # run on the CPU instead, replace the right side with "cpu"
-    __DEVICE = 0 if torch.cuda.is_available() else "cpu"
```
TongguangZhang, line 42: was this line causing an error? if your computer does have cuda, you should try and use it, if not, it will default to cpu

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +94,30 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
         # * conf
         # * device
         # * verbose
-        predictions = ...
+        self.__model.predict(image, conf=0.7, device=self.__DEVICE, verbose=False)
```
TongguangZhang, line 97: if you're not storing or computing the result, why keep it? also note that the the comment above says you should specify the parameter `source` even though you can pass in parameters by position, it's better to make it explicit

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +94,30 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
 
         # Get the Result object
-        prediction = ...
+        prediction = self.__model(image)
```
TongguangZhang, line 100: this is the wrong way to run inference on the model, as this method is deprecated

modules/bootcamp/detect_landing_pad.py
TongguangZhang, line 97: try to follow the template, quick hint, note that it asks you for predictions plural, and then in line 104, prediction singular, take a look at what model.predict() returns and how u can get a singular result out of it

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +94,30 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
 
         # Get the xyxy boxes list from the Boxes object in the Result object
-        boxes_xyxy = ...
+        boxes_xyxy = prediction[0].boxes.xyxy
 
         # Detach the xyxy boxes to make a copy,
         # move the copy into CPU space,
         # and convert to a numpy array
-        boxes_cpu = ...
+        boxes_cpu = boxes_xyxy.cpu().numpy()
```
TongguangZhang, line 112: remember to call `detach()` on this, this removes it from the computational graph
basically this turns off the gradient tracking, which is used in model training for back propagation, since we're running inference here, we don't want to track the gradients

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +94,30 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
 
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
-        # Hint: .shape gets the dimensions of the numpy array
-        # for i in range(0, ...):
-        #     # Create BoundingBox object and append to list
-        #     result, box = ...
+        for result in enumerate(boxes_cpu):
+            bounding_boxes.append(bounding_box.BoundingBox.create(result[1])[1])
```
TongguangZhang, line 117: there's a reason why create returns a boolean flag, it's used to check whether or not the function failed, you don't want to append a box if the creation failed
in fact, you want to return no boxes at all for the image - in data science, we delete an entire data point if part of it is corrupted, instead of returning partial data
in this case, the data point is 1 image, so if some of the bounding boxes failed to create, we should skip the image

TongguangZhang (commented):
reviewed

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -37,12 +37,43 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
...
+    def get_closest_landing_pad(
+        self, report: drone_report.DroneReport, landing_pad_locations: "list[location.Location]"
+    ) -> location.Location | None:
+        """
+        Find the closest landing pad to the drone.
+        """
+        if not landing_pad_locations:
+            return None
+
+        closest_landing_pad: location.Location = landing_pad_locations[0]
```
TongguangZhang, line 61: you still shouldn't initialize this too the first landing pad, since you then compare it to itself, which is redundant
instead, you should initialize the landing pad to null, and initialize the minimum value to infinity

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -37,12 +37,43 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
...
+        for landing_pad in enumerate(landing_pad_locations):
+            if self.get_squared_distance(
+                report.position.location_x,
+                report.position.location_y,
+                landing_pad[1].location_x,
+                landing_pad[1].location_y,
+            ) < self.get_squared_distance(
```
TongguangZhang, line 68: storing the minimum distance also means you don't have up recalculate this every time, which is inefficient

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -37,12 +37,43 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
...
+        self, report: drone_report.DroneReport, landing_pad_locations: "list[location.Location]"
+    ) -> location.Location | None:
+        """
+        Find the closest landing pad to the drone.
+        """
+        if not landing_pad_locations:
+            return None
+
+        closest_landing_pad: location.Location = landing_pad_locations[0]
+        for landing_pad in enumerate(landing_pad_locations):
```
TongguangZhang, line 62: why enumerate? do you ever use the index, or just the object?

modules/bootcamp/detect_landing_pad.py
```diff
@@ -39,6 +35,7 @@ class DetectLandingPad:
     # Chooses the GPU if it exists, otherwise runs on the CPU
     # If you have a CUDA capable GPU but want to force it to
     # run on the CPU instead, replace the right side with "cpu"
+    # __DEVICE = 0 if torch.cuda.is_available() else "cpu"
```
TongguangZhang, line 38: delete the comment - we try to avoid having "dead" code, which is commented out code that never runs, code in an if false (this is a bad example lol), etc.

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +95,34 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
-        boxes_cpu = ...
+        boxes_cpu = boxes_xyxy.detach().cpu().numpy()
 
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
-        # Hint: .shape gets the dimensions of the numpy array
-        # for i in range(0, ...):
-        #     # Create BoundingBox object and append to list
-        #     result, box = ...
+        for boxes in enumerate(boxes_cpu):
```
TongguangZhang, line 119: again, why enumerate?

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +95,34 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+            result, box = bounding_box.BoundingBox.create(boxes[1])
+            if result:
```
TongguangZhang, line 121: reread the last comment here, instead of skipping a failed box, we skip the whole image
if not result:
    return [], image annotated
bounding_boxes.append(box)

TongguangZhang (commented):
reviewed

TongguangZhang (commented):
lgtm!

PR #121, opened 2024-11-07
Bootcamp - Arpan

arpanroy18:
Finished bootcamp

### DIFF
--- a/modules/bootcamp/decision_simple_waypoint.py
+++ b/modules/bootcamp/decision_simple_waypoint.py
@@ -37,36 +37,26 @@ class DecisionSimpleWaypoint(base_decision.BaseDecision):
         # ============
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
-
-        self.command_index = 0
-        self.commands = [
-            commands.Command.create_set_relative_destination_command(
-                self.waypoint.location_x, self.waypoint.location_y
-            ),
-        ]
-
         self.has_sent_destination_command = False
         self.has_sent_landing_command = False
-        
-
         self.counter = 0
 
         # New variables for handling GPS noise
         self.within_radius_count = 0
         self.consecutive_threshold = 3
 
-        def check_radius(
-            self,
-            current_x: location.Location,
-            current_y: location.Location,
-            waypoint_x: location.Location,
-            waypoint_y: location.Location,
-        ) -> bool:
-            """
-            Check if the drone is within the acceptance radius of the waypoint.
-            """
-            radius = math.sqrt((waypoint_x - current_x) ** 2 + (waypoint_y - current_y) ** 2)
-            return radius < self.acceptance_radius
+    def check_radius(
+        self,
+        current_x: location.Location,
+        current_y: location.Location,
+        waypoint_x: location.Location,
+        waypoint_y: location.Location,
+    ) -> bool:
+        """
+        Check if the drone is within the acceptance radius of the waypoint.
+        """
+        radius = math.sqrt((waypoint_x - current_x) ** 2 + (waypoint_y - current_y) ** 2)
+        return radius < self.acceptance_radius
 
         # ============
         # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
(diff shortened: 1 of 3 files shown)


modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,7 +103,61 @@ def run(
...
+                    print("Waypoint reached. Preparing to land at the nearest landing pad.")
+            else:
+                if not self.has_sent_landing_command:
+                    # Find the closest landing pad
+                    closest_pad = self.find_closest_landing_pad(
+                        drone_position, landing_pad_locations
+                    )
+                    if closest_pad is None:
+                        print("No landing pads available. Remaining idle.")
+                        return command  # Remain idle if no landing pads are found
```
AaronWang04, line 138: in the event this happens, its probably better to return with a None so that it can be properly handled

AaronWang04 (commented):
almost good

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -32,12 +34,31 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
...
+            commands.Command.create_set_relative_destination_command(
+                self.waypoint.location_x, self.waypoint.location_y
+            ),
+        ]
+
+        self.has_sent_landing_command = False
+
+        self.counter = 0
+
+        @staticmethod
```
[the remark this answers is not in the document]
AaronWang04: resolve this issue

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,8 +89,31 @@ def run(
...
+        elif (
+            report.status == drone_status.DroneStatus.HALTED
+            and not self.has_sent_landing_command
+            and self.check_radius(
+                report.position.location_x,
+                report.position.location_y,
+                self.waypoint.location_x,
+                self.waypoint.location_y,
+            )
+        ):
```
AaronWang04, line 111: what if the drone doesnt satisfy self.check_radius after the initial command due to some numerical error (that can definitely happen with analog devices such as a gps?

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -21,28 +21,62 @@
...
+        Args:
+            loc1 (location.Location): The first location.
+            loc2 (location.Location): The second location.
+
+        Returns:
+            float: The squared Euclidean distance between loc1 and loc2.
+        """
+        dx = loc2.location_x - loc1.location_x
+        dy = loc2.location_y - loc1.location_y
+        return dx * dx + dy * dy
```
AaronWang04, line 57: good formatting and comments

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,8 +87,42 @@ def run(
...
+                # Increment the counter if within radius
+                self.within_radius_count += 1
+            else:
+                # Reset the counter if outside the radius
+                self.within_radius_count = 0
+
+            # Check if the drone has been within the radius for the required number of checks
+            if self.within_radius_count >= self.consecutive_threshold:
+                command = commands.Command.create_land_command()
+                self.has_sent_landing_command = True
```
AaronWang04, line 122: this doesnt really fix it. think about it like this.
1. you have sent the destination command
2. the drone has finished executing that command BUT is not within acceptance radius of the actual location which can happen due to numerical instability
3. now the drone has self.has_sent_landing_command as true, and within_radius is always 0 because its not within the bounds (unless you hit the numerical error again and GPA says your within the radius)

due to numerical instability, this function will probably work in the real world but is not how its intended to work

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,8 +87,52 @@ def run(
...
+                )
+                self.has_sent_destination_command = True
+
+            # Check if the drone has been within the radius for the required number of checks
+            if (
+                self.within_radius_count >= self.consecutive_threshold
+                and not self.has_sent_landing_command
+            ):
+                command = commands.Command.create_land_command()
+                self.has_sent_landing_command = True
```
AaronWang04, line 132: working, but if you look carefully you essentially don't need the has sent destination command. you only need to check if you are within the radius since if not you would be sending the destination command

AaronWang04 (approved):
approved

PR #124, opened 2024-11-12
Final Version

joeyhlu:


modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +97,38 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
-        #     result, box = ...
-
-        return [], image_annotated
+        for i in range(0, boxes_cpu.shape[0]):
+            #     # Create BoundingBox object and append to list
+            #     result, box = ...
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+            if not result:
+                continue
+            assert box is not None
```
AaronWang04, line 128: generally we return an error value (like None) instead of asserting which is then caught at a layer above this (in the same way how if not result we skip it)

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +97,38 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
-        #     # Create BoundingBox object and append to list
-        #     result, box = ...
-
-        return [], image_annotated
+        for i in range(0, boxes_cpu.shape[0]):
+            #     # Create BoundingBox object and append to list
+            #     result, box = ...
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+            if not result:
+                continue
```
AaronWang04, line 127: the more common practice is to disregard the whole image taken if any of its bounding box has an error (faulty data essentially)

AaronWang04 (approved):
mostly LGTM, some nitpicks at detech landing pad that you can take a look at

PR #128, opened 2024-11-24
Completed Bootcamp

lilyjge:


modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +95,35 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         bounding_boxes = []
         # Hint: .shape gets the dimensions of the numpy array
-        # for i in range(0, ...):
-        #     # Create BoundingBox object and append to list
-        #     result, box = ...
+        for i in range(0, boxes_cpu.shape[0]):
+            #     # Create BoundingBox object and append to list
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+            if result:
+                bounding_boxes.append(box)
```
AaronWang04, line 124: its common practice to disregard the entire inference if there is any result

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,6 +93,40 @@ def run(
...
+                    print(self.waypoint)
+                    print(report.position)
+                    print(self.is_same(self.waypoint, report.position))
+                    command = commands.Command.create_set_relative_destination_command(
+                        self.waypoint.location_x - report.position.location_x,
+                        self.waypoint.location_y - report.position.location_y,
+                    )
+            # At landing pad.
+            else:
+                command = commands.Command.create_land_command()
```
AaronWang04, line 120: the intended solution is to add another is_same check here with the landing pad location. in fact this one matters a lot more than the previous check as you need to be physically landing on a landing pad

modules/bootcamp/detect_landing_pad.py
lilyjge: Could you explain more? Why should the inference be disregarded if there is an result? What's the point of having a list of boxes then? Thank you!

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -69,6 +79,17 @@ def run(
         # ============
 
         # Do something based on the report and the state of this class...
+        if report.status.name == "HALTED":
```
AaronWang04, line 82: I wouldnt use the name here, you can directly associate with class type

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -43,6 +43,16 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
...
 
+    def check_proximity(self, position: location.Location) -> bool:
+        """
+        Returns True if position is within self's acceptance radius of the waypoint.
+        """
+        return (
+            abs(position.location_x - self.waypoint.location_x) ** 2
+            + abs(position.location_y - self.waypoint.location_y) ** 2
+            < self.acceptance_radius**2
+        )
```
AaronWang04, line 54: good job not using sqrt to reduce computation

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -38,11 +38,36 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
         # ============
 
         # Add your own
+        self.closest_landing_pad = location.Location(0, 0)
```
AaronWang04, line 41: initialization with none is probably better convention, the value of location(0,0) will never be used

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,6 +94,40 @@ def run(
...
+                    command = commands.Command.create_set_relative_destination_command(
+                        self.waypoint.location_x - report.position.location_x,
+                        self.waypoint.location_y - report.position.location_y,
+                    )
+            # At landing pad.
+            elif self.is_same(self.closest_landing_pad, report.position):
+                command = commands.Command.create_land_command()
+        elif report.status.name == "MOVING":
+            # At waypoint OR waypoint has been reached and at landing pad.
+            if self.is_same(self.waypoint, report.position) or (
```
AaronWang04, line 124: this wouldnt work as the drone may never leave the waypoint.

the goal is for the drone to arrive at the waypoint, snap a picture to find landing pads, and move to an apprioate landing pad to land on.

I think elif moving: pass is fine here

AaronWang04 (approved):

PR #129, opened 2024-12-16
Completed Autonomy Bootcamp

Baighasan:


modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -70,6 +68,21 @@ def run(
 
         # Do something based on the report and the state of this class...
 
+        # Distance between two points formula
+        distance_from_waypoint = (
+            (self.waypoint.location_x - report.position.location_x) ** 2
+            + (self.waypoint.location_y - report.position.location_y) ** 2
+        ) ** 0.5
+        if distance_from_waypoint < self.acceptance_radius:
+            if report.status.value == 1:
```
maxlou05, line 77: Please use the enum type (like the variable, the words) rather than a "magic number". This enhances code readability.

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -70,6 +68,21 @@ def run(
 
         # Do something based on the report and the state of this class...
 
+        # Distance between two points formula
+        distance_from_waypoint = (
+            (self.waypoint.location_x - report.position.location_x) ** 2
+            + (self.waypoint.location_y - report.position.location_y) ** 2
+        ) ** 0.5
```
maxlou05, line 75: Can you think of a way to avoid using the square root? (Hint: think about the purpose for which you are using this distance)

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +95,34 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+        predictions = self.__model.predict(image)
 
         # Get the Result object
-        prediction = ...
+        prediction = predictions[0]
 
         # Plot the annotated image from the Result object
         # Include the confidence value
-        image_annotated = ...
+        image_annotated = prediction.plot()
```
maxlou05, line 105: Please also include the confidence value in the annotated image

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +95,34 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
         # * conf
         # * device
         # * verbose
-        predictions = ...
+        predictions = self.__model.predict(image)
```
maxlou05, line 98: Please set all of the parameters of interest (in the comments above) explicitly

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +95,34 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
-        # for i in range(0, ...):
-        #     # Create BoundingBox object and append to list
-        #     result, box = ...
-
-        return [], image_annotated
+        dim_box = boxes_cpu.shape
+        for i in range(0, dim_box[0]):
+            # Create BoundingBox object and append to list
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+            assert result
```
maxlou05, line 122: Do not use `assert` here, we do not want throw an error and cause our program to crash mid-flight, which may cause the drone to fall out of the sky. Instead, think of a better way to handle the error

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -19,6 +19,19 @@
...
+def calculate_distance(x1: float, y1: float, x2: float, y2: float) -> float:
+    """
+    Calculates the distance between two given points
+    :param x1:
+    :param y1:
+    :param x2:
+    :param y2:
+    :return:
+    """
+    distance = ((x2 - x1) ** 2 + (y2 - y1) ** 2) ** 0.5
```
maxlou05, line 31: Again, can you think of a way to avoid using the square root operation (it's more expensive)

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,7 +81,49 @@ def run(
...
+            self.waypoint.location_x,
+            self.waypoint.location_y,
+        )
+        if self.__destination_type == "landing_pad":
+            distance_from_pad = calculate_distance(
+                report.position.location_x,
+                report.position.location_y,
+                report.destination.location_x,
+                report.destination.location_y,
+            )
```
maxlou05, line 96: `landing_pad_locations` has a list of `location.Location`, as you can see in the function definition. Use those as the coordinates of possible landing pads. `report.destination` is simply the last location that you requested to go to. I see that you calculated the closest pad, but you will need to save it in a variable somewhere so you can use it here. (Hint: class attributes are saved across `run()` calls!)
btw, in your simulation, you should see your drone land on a landing pad

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,7 +81,49 @@ def run(
...
+            if report.status.value == 1:
+                command = commands.Command.create_land_command()
+            elif distance_from_pad < self.acceptance_radius:
+                command = commands.Command.create_halt_command()
+
+        elif distance_from_waypoint < self.acceptance_radius:
+            if report.status.value == 1:
```
maxlou05, line 103: Please use the enum value

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,7 +81,49 @@ def run(
...
+                min_distance = float("inf")
+                closest_pad = None
+                for l in landing_pad_locations:
```
maxlou05, line 106: Use a better name than `l`. Be more descriptive: don't be afraid of using long variable names! In fact, we encourage it.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,7 +81,49 @@ def run(
...
+
+                command = commands.Command.create_set_relative_destination_command(
+                    closest_pad.location_x - report.position.location_x,
+                    closest_pad.location_y - report.position.location_y,
+                )
+                self.__destination_type = "landing_pad"
+
+            else:
+                command = commands.Command.create_halt_command()
+        elif report.status.value == 1:
```
maxlou05, line 122: Use enum value

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,7 +81,49 @@ def run(
...
+            self.waypoint.location_y,
+        )
+        if self.__destination_type == "landing_pad":
+            distance_from_pad = calculate_distance(
+                report.position.location_x,
+                report.position.location_y,
+                report.destination.location_x,
+                report.destination.location_y,
+            )
+            if report.status.value == 1:
```
maxlou05, line 97: Use enum value

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,7 +81,49 @@ def run(
...
+                command = commands.Command.create_land_command()
```
maxlou05, line 98: Why does it land no matter how close the drone is to the landing pad? I think you should check the distance first, before checking this

maxlou05 (commented):
Reviewed

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -36,9 +36,7 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
         # ============
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
-
-        # Add your own
-
+        self.__halted = 1
```
maxlou05, line 39: I meant `drone_status.DroneStatus.HALTED` as the enum. The `DroneStatus` is the enum

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -70,6 +68,21 @@ def run(
...
+            self.waypoint.location_y - report.position.location_y
+        ) ** 2
+        # Squares the acceptance radius due to lack of square root in the distance calculation
+        if distance_from_waypoint < self.acceptance_radius**2:
+            if report.status.value == self.__halted:
+                command = commands.Command.create_land_command()
+            else:
+                command = commands.Command.create_halt_command()
+        elif report.status.value == self.__halted:
+            command = commands.Command.create_set_relative_destination_command(
```
maxlou05, line 82: This is a relative distance command, your own position matters

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -19,6 +19,20 @@
 # pylint: disable=duplicate-code,unused-argument
 
 
+def calculate_distance(x1: float, y1: float, x2: float, y2: float) -> float:
+    """
+    Calculates the distance between two given points
+    NOTE: distance is not squared
```
maxlou05, line 25: It seems that your distance is squared? You are not taking the square root (I think you just forgot to change the comment)

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -37,7 +51,9 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
-        # Add your own
+        self.__destination_type = ""
+        self.__closest_pad_location = None
+        self.__halted = 1
```
maxlou05, line 56: You can use `drone_status.DroneStatus.HALTED`

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,7 +84,59 @@ def run(
...
+                report.position.location_x,
+                report.position.location_y,
+                self.waypoint.location_x,
+                self.waypoint.location_y,
+            )
+            # Squares the acceptance radius due to lack of square root in the distance calculation
+            if distance_from_waypoint < self.acceptance_radius**2:
+                if report.status.value == self.__halted:
+                    min_distance = float("inf")
+                    closest_pad = None
```
maxlou05, line 112: Instead of doing this, you can just directly use your `self.__closest_pad_location` as the variable, since this is never really used on its own

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +95,37 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
-        #     # Create BoundingBox object and append to list
-        #     result, box = ...
-
-        return [], image_annotated
+        dim_box = boxes_cpu.shape
+        for i in range(0, dim_box[0]):
+            # Create BoundingBox object and append to list
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+            if not result:
+                continue
```
maxlou05, line 125: In statistics, we normally discard the entire data point (the entire image) if a part of the data (a bounding box) is incorrect. (You don't have to change anything here, just something to think about)

maxlou05 (commented):
Reviewed

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,7 +83,59 @@ def run(
...
+            if distance_from_waypoint < self.acceptance_radius**2:
+                if report.status.value == drone_status.DroneStatus.HALTED.value:
+                    min_distance = float("inf")
+                    for landing_pad in landing_pad_locations:
+                        x1, y1 = report.position.location_x, report.position.location_y
+                        x2, y2 = landing_pad.location_x, landing_pad.location_y
+                        distance = calculate_distance(x1, y1, x2, y2)
+                        if distance < min_distance:
+                            min_distance = distance
+                            self.__closest_pad_location = location.Location(x2, y2)
```
maxlou05, line 117: Normally, you would just do `self.__closest_pad_location = landing_pad` here instead of constructing a new object, but that's fine (you do not need to change anything here, just something to think about)

maxlou05 (commented):
Conditionally approved

modules/bootcamp/decision_waypoint_landing_pads.py
Baighasan: Got it 👍

PR #130, opened 2024-12-18
Completed Autonomy Bootcamp

samcyx:
Completed bootcamp

### DIFF
--- a/modules/bootcamp/decision_simple_waypoint.py
+++ b/modules/bootcamp/decision_simple_waypoint.py
@@ -98,9 +98,8 @@ class DecisionSimpleWaypoint(base_decision.BaseDecision):
         ):  # the drone is at the waypoint and stopped
             command = commands.Command.create_land_command()
 
-        elif (
-            current_status.value == drone_status.DroneStatus.MOVING
-            and self._within_radius(current_position)
+        elif current_status.value == drone_status.DroneStatus.MOVING and self._within_radius(
+            current_position
         ):  # necessary? Halt when reach the waypoint
             command = commands.Command.create_halt_command()
 
--- a/modules/bootcamp/decision_simple_waypoint.py
+++ b/modules/bootcamp/decision_simple_waypoint.py
@@ -41,24 +41,26 @@ class DecisionSimpleWaypoint(base_decision.BaseDecision):
         # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
         # ============
 
-    def _distance_between_locations(self, l1: location.Location, l2: location.Location) -> float:
+    def _squared_distance_between_locations(
+        self, l1: location.Location, l2: location.Location
+    ) -> float:
         """
-        Returns the distance between two locations, used by the within_radius function
+        Returns the squared distance between two locations,
+        used to reduce computational overhead for calculating relative distance parity
         """
         x1 = l1.location_x
         x2 = l2.location_x
         y1 = l1.location_y
         y2 = l2.location_y
-        distance = ((x1 - x2) ** 2 + (y1 - y2) ** 2) ** 0.5
+        distance = (x1 - x2) ** 2 + (y1 - y2) ** 2
         return distance
 
     def _within_radius(self, current_location: location.Location) -> bool:
         """
         Returns whether the drone is within the waypoint
         """
-        distance = self._distance_between_locations(current_location, self.waypoint)
-        print(distance, self.acceptance_radius)
-        if distance < self.acceptance_radius:
+        distance = self._squared_distance_between_locations(current_location, self.waypoint)
+        if distance < self.acceptance_radius**2:
             return True
         return False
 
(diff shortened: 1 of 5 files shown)


modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -36,13 +36,32 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
...
 
+    def _distance_between_locations(self, l1: location.Location, l2: location.Location) -> float:
+        """
+        Returns the distance between two locations, used by the within_radius function
+        """
+        x1 = l1.location_x
+        x2 = l2.location_x
+        y1 = l1.location_y
+        y2 = l2.location_y
+        distance = ((x1 - x2) ** 2 + (y1 - y2) ** 2) ** 0.5
```
maxlou05, line 52: Try to avoid using the square root function (it's expensive) (Hint: think about what is the purpose you are using the distance for)

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -69,7 +88,34 @@ def run(
         # ============
 
         # Do something based on the report and the state of this class...
-
+        current_position = report.position
+        current_status = report.status
+        current_destination = report.destination
+        # moving = 0
+        # halted = 1
+        # landed = 2
```
maxlou05, line 96: Instead of doing this, please use the real enum variable. It's more readable and you don't have to write these comments

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -69,7 +88,34 @@ def run(
...
+        if current_status.value == 1 and self._within_radius(
+            current_position
+        ):  # the drone is at the waypoint and stopped
+            command = commands.Command.create_land_command()
+
+        elif (
+            current_status.value == 0 and self.waypoint == current_position
```
maxlou05, line 103: You should use your `_within_radius` function here

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -69,7 +88,34 @@ def run(
...
+        ):  # incorrect final destination or halt location
+            command = commands.Command.create_set_relative_destination_command(
+                self.waypoint.location_x - current_position.location_x,
+                self.waypoint.location_y - current_position.location_y,
+            )
+
+        elif (
+            current_status.value == 0 and current_destination == self.waypoint
+        ) or current_status.value == 2:  # drone is travelling towards waypoint or already landed
+            command = commands.Command.create_null_command()
```
maxlou05, line 118: The default command is the null command (see line 84), so you don't actually have to do this

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -37,12 +37,61 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
...
         # ============
 
-        # Add your own
+        self.land = False
 
         # ============
         # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
         # ============
 
+    def _distance_between_locations(self, l1: location.Location, l2: location.Location) -> float:
```
maxlou05, line 46: You shouldn't need this, think of how you can avoid using this

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -37,12 +37,61 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
...
+        print(distance, self.acceptance_radius)
+        if distance < self.acceptance_radius:
+            return True
+        return False
+
+    def _find_landing_pad(
+        self, current_location: location.Location, landing_pad_locations: "list[location.Location]"
+    ) -> location.Location:
+        closest_landing_pad = landing_pad_locations[0]
+        for i in landing_pad_locations[1:]:
```
maxlou05, line 85: Great use of list splicing here! Although, what would happen if the landing pad locations was empty? (You do not need to change anything here, just something to think about)

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,8 +117,39 @@ def run(
...
         # ============
 
-        # Do something based on the report and the state of this class...
-
+        current_position = report.position
+        current_status = report.status
+        current_destination = report.destination
+        # moving = 0
+        # halted = 1
+        # landed = 2
```
maxlou05, line 125: Again, please use enum

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,8 +117,39 @@ def run(
...
+            current_position
+        ):  # the drone is at the waypoint and stopped
+            closest_landing_pad = self._find_landing_pad(current_position, landing_pad_locations)
+            self.waypoint = closest_landing_pad
+            if self.land:
+                command = commands.Command.create_land_command()
+            else:
+                self.land = True
+        elif (
+            current_status.value == 0 and self.waypoint == current_position
```
maxlou05, line 137: Use your `_within_radius` function here. It is not great to compare floating point numbers directly

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,8 +117,39 @@ def run(
...
+        ):  # incorrect final destination or halt location
+            command = commands.Command.create_set_relative_destination_command(
+                self.waypoint.location_x - current_position.location_x,
+                self.waypoint.location_y - current_position.location_y,
+            )
+
+        elif (
+            current_status.value == 0 and current_destination == self.waypoint
+        ) or current_status.value == 2:  # drone is travelling towards waypoint or already landed
+            command = commands.Command.create_null_command()
```
maxlou05, line 152: Again, default command is null, so you don't actually have to do this

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,8 +117,39 @@ def run(
...
+
+        if current_status.value == 1 and self._within_radius(
+            current_position
+        ):  # the drone is at the waypoint and stopped
+            closest_landing_pad = self._find_landing_pad(current_position, landing_pad_locations)
+            self.waypoint = closest_landing_pad
+            if self.land:
+                command = commands.Command.create_land_command()
+            else:
+                self.land = True
```
maxlou05, line 135: Although this is ok, it would be nice to have some indicator variable to tell when the drone has landed at the landing pad. (You do not need to change anything here, just something to think about)

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +96,41 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+        )
 
         # Get the Result object
-        prediction = ...
+        prediction = predictions[0]
 
         # Plot the annotated image from the Result object
         # Include the confidence value
-        image_annotated = ...
+        image_annotated = prediction.plot()
```
maxlou05, line 108: Please include the confidence value in the annotated image

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +96,41 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Hint: .shape gets the dimensions of the numpy array
         # for i in range(0, ...):
         #     # Create BoundingBox object and append to list
         #     result, box = ...
-
-        return [], image_annotated
+        for i in range(0, boxes_cpu.shape[0]):
+            b_box = bounding_box.BoundingBox.create(boxes_cpu[i])
+            bounding_boxes.append(b_box[1])
+            if not b_box[0]:
```
maxlou05, line 131: You should probably check this first, before you append the result to your list, although it doesn't really matter in your case. Also, please use tuple unpacking: `var1, var2 = BoundingBox.create()`

maxlou05 (commented):
Reviewed

maxlou05 (commented):
Approved!

PR #131, opened 2024-12-20
Gurvir - WARG Autonomy Bootcamp

Gurvirr:


modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +98,39 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Hint: .shape gets the dimensions of the numpy array
         # for i in range(0, ...):
         #     # Create BoundingBox object and append to list
         #     result, box = ...
 
-        return [], image_annotated
+        for i in range(boxes_cpu.shape[0]):
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+            if result:
+                bounding_boxes.append(box)
```
maxlou05, line 131: Normally in statistics, if one part (one of the bounding boxes) of the data (the entire image) is invalid, we discard the entire data point.
(You do not need to change anything here, just something to think about)

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,6 +78,29 @@ def run(
...
+        distance_y = self.destination_y - location_y
+
+        distance_squared = distance_x**2 + distance_y**2
+        acceptance_radius_squared = self.acceptance_radius**2
+
+        # Check if the drone should land or move toward the waypoint
+        if not self.has_sent_landing_command and report.status == drone_status.DroneStatus.HALTED:
+
+            # If the drone is halted, start the move command
+            command = self.commands
```
maxlou05, line 96: I don't think `self.commands` is defined anywhere

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,6 +78,29 @@ def run(
...
+            self.has_sent_landing_command = True
```
maxlou05, line 97: Please rename this variable to `has_sent_move_command`, because your comments indicate that you sent a move command and not a landing command. It's very confusing that you're setting `has_sent_landing_command` to True

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -39,6 +39,16 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
...
 
+        self.has_sent_landing_command = False
+
+        # Destination coords
+        self.destination_x = self.waypoint.location_x
+        self.destination_y = self.waypoint.location_y
+
+        self.move_command = commands.Command.create_set_relative_destination_command(
+            self.destination_x, self.waypoint.location_y
+        )
```
maxlou05, line 50: Not sure what this is for? It is never used. Also, this move command is based on relative distance, so you need to take into account your current position

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -70,6 +72,51 @@ def run(
...
+                    self.closest_landing_pad = pad
+
+            if self.closest_landing_pad:
+                command = commands.Command.create_set_relative_destination_command(
+                    self.closest_landing_pad.location_x - report.position.location_x,
+                    self.closest_landing_pad.location_y - report.position.location_y,
+                )
+
+        # If the drone is at the closest landing pad, issue the land command
+        elif (
```
maxlou05, line 112: This structure works well in the bootcamp, where the commands cannot get interrupted. What happens if you picked a closest landing pad, and then your drone had to emergency stop and reroute due to an obstacle? What would it do after that? Your current code will have it stay there and do nothing

maxlou05 (commented):
Reviewed

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,6 +70,30 @@ def run(
...
+        acceptance_radius_squared = self.acceptance_radius**2
+
+        # Check if the drone should land or move toward the waypoint
+        if not self.has_sent_move_command and report.status == drone_status.DroneStatus.HALTED:
+
+            # If the drone is halted, start the move command
+            command = move_command
+            self.has_sent_move_command = True
+
+        elif distance_squared < acceptance_radius_squared and self.has_sent_move_command is True:
```
maxlou05, line 92: This is fine for this bootcamp, but what happens if the drone halts unexpectedly midway in the flight? (Just something to think about, you do not need to change anything)

maxlou05 (commented):
Conditionally approved!

PR #136, opened 2025-01-07
Bootcamp Pull

RohanKatreddy:


modules/bootcamp/detect_landing_pad.py
```diff
@@ -122,7 +122,10 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
         #     # Create BoundingBox object and append to list
         #     result, box = ...
 
-        return [], image_annotated
+        for i in range(0, boxes_cpu.shape[0]):
+            bounding_boxes.append(bounding_box.BoundingBox.create(boxes_cpu[i])[1])
+                        
+        return bounding_boxes, image_annotated
```
TongguangZhang, line 128: take a closer look at the create() function, there's a reason why it returns a tuple, take a look at what the first element of the tuple represents

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,22 +98,22 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
 
         # Get the xyxy boxes list from the Boxes object in the Result object
-        boxes_xyxy = ...
+        boxes_xyxy = prediction.boxes.xyxy
 
         # Detach the xyxy boxes to make a copy,
         # move the copy into CPU space,
         # and convert to a numpy array
-        boxes_cpu = ...
+        boxes_cpu = boxes_xyxy.cpu().numpy()
```
TongguangZhang, line 116: make sure to call detach(), this removes it from the computational graph
basically, these tensors have gradient tracking, which is used for back propagation in model training, but since we're using the inference, we don't need the gradients, and they take up a lot of extra memory

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,22 +98,22 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
         # * conf
         # * device
         # * verbose
-        predictions = ...
+        predictions = self.__model.predict(image, conf=0.25, device=self.__DEVICE, verbose=True)
```
TongguangZhang, line 101: 0.25 might be a little too low here

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -67,8 +67,25 @@ def run(
...
         # ============
-
-        # Do something based on the report and the state of this class...
+        def get_x_difference(initial: location.Location, final: location.Location):
+            return (final.location_x - initial.location_x)
+        
+        def get_y_difference(initial: location.Location, final: location.Location):
+            return (final.location_y - initial.location_y)
+        
+        def get_distance_squared(diff_x, diff_y):
```
TongguangZhang, line 76: let's use **2 to make it more readable that ur doing the square

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -67,8 +67,25 @@ def run(
...
+            return (diff_x*diff_x + diff_y*diff_y)
+        
+        x_difference = get_x_difference(report.position, self.waypoint)
+        y_difference = get_y_difference(report.position, self.waypoint)
+
+        
+        if (report.status == drone_status.DroneStatus.MOVING):
+            command = commands.Command.create_null_command()
```
TongguangZhang, line 84: do you need to do this? if you remove the if statement and the drone is moving, wouldn't it return a null command by default?

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,7 +68,41 @@ def run(
...
         # ============
 
-        # Do something based on the report and the state of this class...
+        def get_x_difference(initial: location.Location, final: location.Location):
+            return (final.location_x - initial.location_x)
+        
+        def get_y_difference(initial: location.Location, final: location.Location):
+            return (final.location_y - initial.location_y)
+        
+        def get_distance_squared(diff_x, diff_y):
```
TongguangZhang, line 77: same as above, use **2

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,7 +68,41 @@ def run(
...
+            return (diff_x*diff_x + diff_y*diff_y)
+        
+        def get_closest_landing_pad():
+            min_distance = 1e20
```
TongguangZhang, line 81: instead of using a very large number, try looking into how u represent infinity in python

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,7 +68,41 @@ def run(
...
+            closest_landing_pad = None
```
TongguangZhang, line 82: make sure you handle the case where the landing pad is still none (no landing pads in the list)

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,7 +68,41 @@ def run(
...
+            return closest_landing_pad
+
+        if self.reached_waypoint:
+            self.waypoint = get_closest_landing_pad()
+
+        x_difference = get_x_difference(report.position, self.waypoint)
+        y_difference = get_y_difference(report.position, self.waypoint)
+        
+        if (report.status == drone_status.DroneStatus.MOVING):
+            command = commands.Command.create_null_command()
```
TongguangZhang, line 98: same as above, is this necessary?

TongguangZhang (commented):
reviewed

modules/bootcamp/decision_simple_waypoint.py
RohanKatreddy: If I remove the if statement the drone gets spammed with set_destination commands since most of the time its quite a ways from the landing pad. I might be missing something but thats the reason that I added it

modules/bootcamp/decision_waypoint_landing_pads.py
RohanKatreddy: If I remove the if statement the drone gets spammed with set_destination commands since most of the time its quite a ways from the landing pad. I might be missing something but thats the reason that I added it

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,7 +97,25 @@ def run(
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
-        # Do something based on the report and the state of this class...
+        if self.reached_waypoint:
+            result = self.get_closest_landing_pad(report, landing_pad_locations)
+            # No landing pads found => land at current location.
+            self.waypoint = result if result is not None else report.position
```
TongguangZhang, line 103: this is incorrect, the instructions don't say to land at the waypoint if no landing pads are found, and you shouldn't make that assumption
since it doesn't tell you how to handle the case, you should return the default command if there are no landing pads and let the rest of the code handle it (consider in a larger system, we might decide to fly around and search for landing pads if none are found, not land)

TongguangZhang, line 103: it's a little bit confusing to overwrite self.waypoint with the landing pad's location, I would store a self.landing_pad variable, but it's fine to do it this way too, you don't need to change anything here

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,7 +97,25 @@ def run(
...
+
+        x_difference = self.get_x_difference(report.position, self.waypoint)
+        y_difference = self.get_y_difference(report.position, self.waypoint)
+
+        if report.status == drone_status.DroneStatus.MOVING:
+            command = commands.Command.create_null_command()
+        elif self.get_distance_squared(x_difference, y_difference) >= (self.acceptance_radius**2):
```
TongguangZhang, line 110: same as above, check drone status is halted

TongguangZhang (commented):
reviewed

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,7 +82,22 @@ def run(
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
-        # Do something based on the report and the state of this class...
+        x_difference = DecisionSimpleWaypoint.get_x_difference(report.position, self.waypoint)
+        y_difference = DecisionSimpleWaypoint.get_y_difference(report.position, self.waypoint)
+
+        if report.status in {
+            drone_status.DroneStatus.MOVING,
+            report.status == drone_status.DroneStatus.LANDED,
```
TongguangZhang, line 90: this is incorrect, remove the report.status == - this just evaluates to True or False, so the whole statement becomes report.status in {MOVING, True/False}, which is not what you want

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -37,12 +32,49 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
...
+        return final.location_y - initial.location_y
+
+    @staticmethod
+    def get_distance_squared(diff_x: float, diff_y: float) -> float:
+        """
+        Return the distance squared of two locations based on their difference in x and y values.
+        """
+        return diff_x**2 + diff_y**2
+
+    def get_closest_landing_pad(
```
TongguangZhang, line 62: this can also be static - the functions you call here are static methods, which don't require an instance of the class (self) to call

TongguangZhang (commented):
reviewed

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,7 +100,30 @@ def run(
...
+        if self.reached_waypoint:
+            self.waypoint = self.get_closest_landing_pad(report, landing_pad_locations)
+
+        x_difference = DecisionWaypointLandingPads.get_x_difference(report.position, self.waypoint)
+        y_difference = DecisionWaypointLandingPads.get_y_difference(report.position, self.waypoint)
+
+        # If self.waypoint is None (also handles the case when no landing pad is found) default null command is sent
+        if (
+            report.status
+            in {drone_status.DroneStatus.MOVING, report.status == drone_status.DroneStatus.LANDED}
```
TongguangZhang, line 112: same mistake here

PR #137, opened 2025-01-08
Bill Bootcamp

szbillyang:


maxlou05 (commented):
Conditionally approved.
This code works, but you are often recalculalting the distance repeatedly when it's not necessary (ie when the drone is still moving). You can optimize this by checking the state of the drone first before you do anything.

PR #139, opened 2025-01-08
Khalil Hawari Bootcamp

Khalil-Hawari:
(Awaiting review, as of this post)


modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -61,17 +64,94 @@ def run(
...
+                command = commands.Command.create_set_relative_destination_command(
+                    report.destination.location_x - report.position.location_x,
+                    report.destination.location_y - report.position.location_y,
+                )
+
+            elif not self.reached_wp:
+                command = commands.Command.create_set_relative_destination_command(
+                    self.waypoint.location_x, self.waypoint.location_y
+                )
+                self.reached_wp = True
```
TongguangZhang, line 89: you should name this variable approaching_wp instead

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -61,17 +64,94 @@ def run(
...
+                print("APPROACHING WP:  ", self.waypoint.location_x, self.waypoint.location_y)
+
+            elif not self.reached_lp:
+                self.target_lp = landing_pad_locations[self.find_closest_lp(landing_pad_locations)]
+                lp_relative_x, lp_relative_y = self.relative_target(self.target_lp, report)
+                command = commands.Command.create_set_relative_destination_command(
+                    lp_relative_x, lp_relative_y
+                )
+                self.reached_lp = True
```
TongguangZhang, line 98: similarly, you should name this approaching_lp

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -61,17 +64,94 @@ def run(
...
+            else:
+                command = commands.Command.create_land_command()
+                print("LANDING")
 
-        # Do something based on the report and the state of this class...
+        return command
 
-        # ============
-        # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
-        # ============
```
TongguangZhang, line 75: don't delete the comment guards

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -61,17 +64,94 @@ def run(
...
+    def find_closest_lp(self, landing_pad_locations: "list[location.Location]") -> int:
+        """
+        Given a list of landing pads,
+        determines the index of the closest one to the waypoint.
+        """
+        if len(landing_pad_locations) in [0, 1]:
+            return 0
```
TongguangZhang, line 113: if the list is len(0) here, and you return 0, when you index it above, you'll get an error

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -61,17 +64,94 @@ def run(
...
 
-        return command
+        wp_x = self.waypoint.location_x
+        wp_y = self.waypoint.location_y
+
+        i = 0
+        winning_i = 0
```
TongguangZhang, line 119: it's better to store the landing pad object, not the index of the winning landing pad

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -61,17 +64,94 @@ def run(
...
+        """
+        Given a list of landing pads,
+        determines the index of the closest one to the waypoint.
+        """
+        if len(landing_pad_locations) in [0, 1]:
+            return 0
 
-        return command
+        wp_x = self.waypoint.location_x
+        wp_y = self.waypoint.location_y
```
TongguangZhang, line 116: why would you use the waypoint location instead of the drone's position?

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -61,17 +64,94 @@ def run(
...
+            same_loc = (
+                wp_x - self.acceptance_radius <= lp_x <= wp_x + self.acceptance_radius
+            ) and (wp_y - self.acceptance_radius <= lp_y <= wp_y + self.acceptance_radius)
+
+            if same_loc:
+                return i
+
+            sq_dist = (lp_x - wp_x) ** 2 + (lp_y - wp_y) ** 2
+
+            if (sq_dist < winning_dist) or (winning_dist == -1):
```
TongguangZhang, line 135: if you wanted to check if it was the first landing pad, I would suggest i == 1
however, we usually want to avoid this kind of hard coded edge case handling, instead, initialize winning_dist to infinity (search up how infinity is represented in python)

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -61,17 +64,94 @@ def run(
...
+        winning_i = 0
+        winning_dist = -1
+
+        for lp in landing_pad_locations:
+            lp_x = lp.location_x
+            lp_y = lp.location_y
+
+            same_loc = (
+                wp_x - self.acceptance_radius <= lp_x <= wp_x + self.acceptance_radius
+            ) and (wp_y - self.acceptance_radius <= lp_y <= wp_y + self.acceptance_radius)
```
TongguangZhang, line 128: why only - and not + this time?
but also, this is again a square, we want to use a circle instead

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -61,17 +64,94 @@ def run(
...
+
+            if same_loc:
+                return i
```
TongguangZhang, line 131: why would you need to explicitly check if it is the same loc? if it was the same loc, wouldn't that landing pad also have the smallest distance and you'd return it anyways?

TongguangZhang (commented):
reviewed

modules/bootcamp/decision_waypoint_landing_pads.py
Khalil-Hawari: Correct, I had figured that if it were the same location then there was no point in continuing the loop -> it could just return right there. 
However, if this rarely ever happens then it's not worth checking. I'll remove it, thx

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -70,6 +70,19 @@ def run(
...
+            delta_y = report.position.location_y - self.waypoint.location_y
+            on_waypoint = delta_x**2 + delta_y**2 < self.acceptance_radius**2
+
+            if on_waypoint:
+                command = commands.Command.create_land_command()
+
+            else:
+                command = commands.Command.create_set_relative_destination_command(
+                    self.waypoint.location_x, self.waypoint.location_y
+                )
```
TongguangZhang, line 84: note that this is relative position, not absolute, you can't just use the location of the waypoint (make sure you run your tests before requesting a rereview

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -61,17 +65,105 @@ def run(
...
         """
-        # Default command
-        command = commands.Command.create_null_command()
-
         # ============
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
-        # Do something based on the report and the state of this class...
+        # print("RUN")
```
TongguangZhang, line 72: it's ok to have print statements, but we want to delete commented out code, take a look at the style guide section on dead code

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -61,17 +65,105 @@ def run(
             put_output(command)
         ```
         """
-        # Default command
-        command = commands.Command.create_null_command()
-
```
TongguangZhang, line 66: don't modify code outside the comment guards - you deleted this and then copied it back down below

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -61,17 +65,105 @@ def run(
...
-        # Do something based on the report and the state of this class...
+        # print("RUN")
+        # Default command
+        command = commands.Command.create_null_command()
+
+        if report.status == drone_status.DroneStatus.HALTED:
+            print("HALTED AT:   ", report.position.location_x, report.position.location_y)
+            print(
+                "DESTINATION WAS:   ", report.destination.location_x, report.destination.location_y
+            )
```
TongguangZhang, line 80: it's worth taking a look at python string formatting - I'd recommend looking at f-string docs, as well as special characters such as -t (no need to change anything here, just smth to keep in mind)

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -61,17 +65,105 @@ def run(
...
+                print("^^UNEXPECTED HALT")
+                command = commands.Command.create_set_relative_destination_command(
+                    report.destination.location_x - report.position.location_x,
+                    report.destination.location_y - report.position.location_y,
+                )
+
+            elif not self.approaching_wp:
+                command = commands.Command.create_set_relative_destination_command(
+                    self.waypoint.location_x, self.waypoint.location_y
+                )
```
TongguangZhang, line 97: incorrect, you need relative instead of absolute coords

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -61,17 +65,105 @@ def run(
...
+        determines the index of the closest one to the waypoint.
+        """
+        if len(landing_pad_locations) == 1:
+            return landing_pad_locations[0]
+
+        drone_x = report.position.location_x
+        drone_y = report.position.location_y
+
+        i = 0
+        winning_pad = landing_pad_locations[0]
```
TongguangZhang, line 135: this still doesn't cover the case where the list is empty - i would suggest initializing winning_pad to None and handling that in the return - if the returned landing pad is none, return the default null command

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -61,17 +65,105 @@ def run(
...
+        Given a list of landing pads,
+        determines the index of the closest one to the waypoint.
+        """
+        if len(landing_pad_locations) == 1:
+            return landing_pad_locations[0]
+
+        drone_x = report.position.location_x
+        drone_y = report.position.location_y
+
+        i = 0
```
TongguangZhang, line 134: if you don't use it, don't declare it

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -61,17 +65,105 @@ def run(
...
+            delta_x = lp_x - drone_x
+            delta_y = lp_y - drone_y
+
+            sq_dist = delta_x**2 + delta_y**2
+
+            if sq_dist < winning_sq_dist:
+                winning_sq_dist = sq_dist
+                winning_pad = landing_pad_locations[i]
+
+            i += 1
```
TongguangZhang, line 151: from above, it looks like this isn't used

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -61,17 +65,105 @@ def run(
...
+
+        # print("TARGET LP FOUND, INDEX: ", winning_i)
+        # print("^^SQDIST FROM WP:  ", winning_dist)
```
TongguangZhang, line 154: delete commented out code, or uncomment it

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +94,46 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
         # * conf
         # * device
         # * verbose
-        predictions = ...
+        predictions = self.__model.predict(
+            source=image, conf=0.7, device=self.__DEVICE, verbose=False
+        )
+        # print("PREDICTIONSSS:  ", prediction)
```
TongguangZhang, line 100: delete or uncomment

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +94,46 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
 
         # Get the Result object
-        prediction = ...
+        prediction = predictions[0]
+        # print("PREDICTION:  ", prediction)
```
TongguangZhang, line 104: delete or uncomment

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +94,46 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Detach the xyxy boxes to make a copy,
-        # move the copy into CPU space,
-        # and convert to a numpy array
-        boxes_cpu = ...
+        # move the copy into CPU space, --> cuda
+        # and convert to a numpy array --> numpy
+        boxes_cpu = boxes_xyxy.detach().cpu().numpy()
+        # print("BOXES-CPU TYPE:  ", type(boxes_cpu))
+        # print("BOXES-CPU:  ", boxes_cpu)
+        # print("BOXES-CPU ND SHAPE", boxes_cpu.shape)
```
TongguangZhang, line 119: delete or uncomment

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +94,46 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
-        #     result, box = ...
 
-        return [], image_annotated
+        # Hint: .shape gets the dimensions of the numpy array
+        for i in range(0, boxes_cpu.shape[0]):
+            # Create BoundingBox object and append to list
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+            # print("RESULT:    ", result)
+            # print("BOX:    ", box)
+
```
TongguangZhang, line 130: delete or uncomment

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +94,46 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+            if result:
+                bounding_boxes.append(box)
+            else:
+                return []
```
TongguangZhang, line 134: is this what the function is supposed to return? take another look at the return annotation, as well as your own code on line 136

TongguangZhang (commented):
reviewed

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -1,77 +1,157 @@
...
+            delta_x = report.position.location_x - lp.location_x
+            delta_y = report.position.location_y - lp.location_y
+
+            sq_dist = delta_x**2 + delta_y**2
+
+            if sq_dist < winning_sq_dist:
+                winning_sq_dist = sq_dist
+                winning_pad = landing_pad_locations[i]
+
+            i += 1
```
maxlou05, line 155: This works, but you also could have just done winning_pad = lp, it's exactly the same thing (no need for the i variable)

PR #140, opened 2025-01-09
Matias' Pull Request

hiatus770:


modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +98,39 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+
         # Hint: .shape gets the dimensions of the numpy array
         # for i in range(0, ...):
         #     # Create BoundingBox object and append to list
         #     result, box = ...
 
-        return [], image_annotated
+        for i in range(boxes_cpu.shape[0]):
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+            if result:
```
TongguangZhang, line 130: instead of skipping failed boxes, we acc wanna return an empty list - in data science, when part of a data point is corrupted, we skip the whole data point instead of returning partial data, in this case, 1 image is 1 data point, and the boxes are part of the image, so we skip the whole image if any of the boxes fail

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,6 +70,24 @@ def run(
...
         # ============
 
+        x_dist = self.waypoint.location_x - report.position.location_x
+        y_dist = self.waypoint.location_y - report.position.location_y
+        dist = (x_dist * x_dist + y_dist * y_dist) ** 0.5
+
+        if report.status == drone_status.DroneStatus.HALTED and dist > self.acceptance_radius:
+            # we are currently stopped so check the next closest waypoint to our destination
+            x_delta = min(max(-60, x_dist), 60)
+            y_delta = min(max(-60, y_dist), 60)
```
TongguangZhang, line 80: this would make the drone fly to the bounds if the waypoint ins outside the bounds, this is not in the instructions and you shouldn't assume it's the intended behaviour
instead, if the waypoint is outside the bounds, you should do nothing (return default command) and leave it for the rest of the system to handle

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,6 +70,24 @@ def run(
...
+            command = commands.Command.create_set_relative_destination_command(x_delta, y_delta)
+            # command = commands.command.create_set_relative_destination_command(x_delta, y_delta)
```
TongguangZhang, line 82: remove dead code - this is commented out code that isn't run

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,6 +70,24 @@ def run(
...
+
+        if dist <= self.acceptance_radius and not self.exited:
+            command = commands.Command.create_land_command()
+            # self.exited = true
```
TongguangZhang, line 86: delete dead code, also, if you don't use self.exited anymore, you should delete it from the rest of your code

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,6 +70,24 @@ def run(
...
+        dist = (x_dist * x_dist + y_dist * y_dist) ** 0.5
+
+        if report.status == drone_status.DroneStatus.HALTED and dist > self.acceptance_radius:
+            # we are currently stopped so check the next closest waypoint to our destination
+            x_delta = min(max(-60, x_dist), 60)
+            y_delta = min(max(-60, y_dist), 60)
+            command = commands.Command.create_set_relative_destination_command(x_delta, y_delta)
+            # command = commands.command.create_set_relative_destination_command(x_delta, y_delta)
+
+        if dist <= self.acceptance_radius and not self.exited:
```
TongguangZhang, line 84: we usually consider acceptance radius to be strictly <, not <=
also, you already checked if the distance is greater above, no need to check if it's less than here

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,6 +69,45 @@ def run(
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
+        x_dist = 0
+        y_dist = 0
+        dist = 10000
```
TongguangZhang, line 74: it's bad practice to use a very large number, instead, take a look at how infinity is represented in python
or you can use the max distance of the bounds

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,6 +69,45 @@ def run(
...
+
+        if self.on_waypoint is False:
```
TongguangZhang, line 76: generally more pythonic to use `if not self.on_waypoint`

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,6 +69,45 @@ def run(
...
+            x_dist = self.waypoint.location_x - report.position.location_x
+            y_dist = self.waypoint.location_y - report.position.location_y
+            dist = (x_dist * x_dist + y_dist * y_dist) ** 0.5
```
TongguangZhang, line 79: use ** 2 instead of x_dist * x_dist

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,6 +70,24 @@ def run(
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
+        x_dist = self.waypoint.location_x - report.position.location_x
+        y_dist = self.waypoint.location_y - report.position.location_y
+        dist = (x_dist * x_dist + y_dist * y_dist) ** 0.5
```
TongguangZhang, line 75: use ** 2 instead of x_dist * x_dist

TongguangZhang, line 75: consider if you need ** 0.5 here - square root is a much more expensive operation than square, maybe you can square the acceptance radius instead?

modules/bootcamp/decision_waypoint_landing_pads.py
TongguangZhang, line 79: consider if you need ** 0.5 here - square root is a much more expensive operation than square, maybe you can square the acceptance radius instead?

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,6 +69,45 @@ def run(
...
+
+        if self.on_waypoint is True:
```
TongguangZhang, line 81: more pythonic to do `if self.on_waypoint`

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,6 +69,45 @@ def run(
...
+            print("waypoint reached")
+            x_dist_set = landing_pad_locations[0].location_x - report.position.location_x
+            y_dist_set = landing_pad_locations[0].location_y - report.position.location_y
```
TongguangZhang, line 84: do not initialize it to the first landing pad
1. the list might be empty - you'd get an index out of range error
2. you're making a redundant calculation since in the loop, you calculate the first landing pad twice

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,6 +69,45 @@ def run(
...
+            dist = (x_dist_set * x_dist_set + y_dist_set * y_dist_set) ** 0.5
```
TongguangZhang, line 85: can we name this min_dist? you use the same variable dist to represent a lot of things

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,6 +69,45 @@ def run(
...
+
+            for landing_pad in landing_pad_locations:
+                x_dist_temp = landing_pad.location_x - report.position.location_x
+                y_dist_temp = landing_pad.location_y - report.position.location_y
```
TongguangZhang, line 89: instead of temp, name it what it is - distance to the landing pad, maybe x_dist_lp

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,6 +69,45 @@ def run(
...
+
+                if (x_dist_temp * x_dist_temp + y_dist_temp * y_dist_temp) ** 0.5 <= dist:
```
TongguangZhang, line 91: use ** 2, avoid using ** 0.5 - consider that if a < b, can we guarantee that a ^ 2 < b ^ 2 if we know a and b are positive?

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,6 +69,45 @@ def run(
...
+                    x_dist_set = x_dist_temp
+                    y_dist_set = y_dist_temp
+
+            x_dist = x_dist_set
+            y_dist = y_dist_set
+            dist = (x_dist_set * x_dist_set + y_dist_set * y_dist_set) ** 0.5
```
TongguangZhang, line 97: you only update the dist if x_dist_set and y_dist_set change, so move it inside the if_statement

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,6 +69,45 @@ def run(
...
+            command = commands.Command.create_land_command()
+            return command
+
+        if dist <= self.acceptance_radius and self.on_waypoint is False:
+            self.on_waypoint = True
+
+        if report.status == drone_status.DroneStatus.HALTED and dist > self.acceptance_radius:
+            # we are currently stopped so check the next closest waypoint to our destination
+            x_delta = min(max(-60, x_dist), 60)
+            y_delta = min(max(-60, y_dist), 60)
```
TongguangZhang, line 109: same as above, don't move to the bounds, return default command

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,6 +69,45 @@ def run(
...
+
+                if (x_dist_temp * x_dist_temp + y_dist_temp * y_dist_temp) ** 0.5 <= dist:
+                    x_dist_set = x_dist_temp
+                    y_dist_set = y_dist_temp
+
+            x_dist = x_dist_set
+            y_dist = y_dist_set
+            dist = (x_dist_set * x_dist_set + y_dist_set * y_dist_set) ** 0.5
+
+        if dist <= self.acceptance_radius and self.on_waypoint is True:
```
TongguangZhang, line 99: this sounds like "land on the waypoint", maybe rename on_waypoint to reached_waypoint for your code to be more readable, just something to consider, you don't have to change anything here

TongguangZhang (commented):
reviewed

AaronWang04 (commented):
some minor style issues

AaronWang04 (commented):
approved

PR #141, opened 2025-01-10
Victor Huang Bootcamp

vichua2006:


modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,7 +68,15 @@ def run(
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
-        # Do something based on the report and the state of this class...
+        # unpack report attributes
+        status = report.status
+        if status == drone_status.DroneStatus.HALTED and self.been_halted_once:
+            command = command.create_land_command()
+        elif status == drone_status.DroneStatus.HALTED:
+            self.been_halted_once = True
```
TongguangZhang, line 76: can you guarantee that the drone is halted at the waypoint? what if the drone halts before that for some reason?

hint: make sure to use the acceptance radius

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,7 +68,15 @@ def run(
...
+            command = command.create_set_relative_destination_command(
+                self.waypoint.location_x, self.waypoint.location_y
+            )
```
TongguangZhang, line 79: note that it's relative destination, not absolute, so you can't just use the waypoint's coordinates

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,7 +69,52 @@ def run(
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
-        # Do something based on the report and the state of this class...
+        # helper functions
+        def find_squared_dist(loc1: location.Location, loc2: location.Location) -> int:
+            """helper function that finds the squared distance between two Location instances"""
+            return (
+                abs(loc1.location_x - loc2.location_x) ** 2
+                + abs(loc1.location_y - loc2.location_y) ** 2
```
TongguangZhang, line 77: do u need abs if ur alr squaring it?

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,7 +69,52 @@ def run(
...
+            )
+
+        def find_closest_location(
+            curr_loc: location.Location, loclist: "list[location.Location]"
+        ) -> location.Location:
+            """find the closest location in a list of locations"""
+            min_dist, res = float("inf"), None
```
TongguangZhang, line 84: you should name the var smth more descriptive than res

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,7 +69,52 @@ def run(
...
+
+            for l in loclist:
```
TongguangZhang, line 86: we use snake_case - take a look at the python style guide in confluence for these kinds of conventions

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,7 +69,52 @@ def run(
...
-        # Do something based on the report and the state of this class...
+        # helper functions
+        def find_squared_dist(loc1: location.Location, loc2: location.Location) -> int:
+            """helper function that finds the squared distance between two Location instances"""
+            return (
+                abs(loc1.location_x - loc2.location_x) ** 2
+                + abs(loc1.location_y - loc2.location_y) ** 2
+            )
+
+        def find_closest_location(
```
TongguangZhang, line 80: you should declare these functions outside of the run function (it's ok to go outside the comment guards in this case)

if you declare it like this, each time the function is called, you redeclare these functions, which is unnecessary

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,7 +69,52 @@ def run(
...
+            final_destination = find_closest_location(position, landing_pad_locations)
+            command = command.create_set_relative_destination_command(
+                final_destination.location_x - position.location_x,
+                final_destination.location_y - position.location_y,
+            )
+        elif status == drone_status.DroneStatus.HALTED:
+            self.been_halted_once = True
+            command = command.create_set_relative_destination_command(
+                self.waypoint.location_x, self.waypoint.location_y
+            )
```
TongguangZhang, line 117: again, relative distance, not absolute

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +95,35 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
         # * conf
         # * device
         # * verbose
-        predictions = ...
+        predictions = self.__model.predict(image, verbose=False)
```
TongguangZhang, line 98: note the other params listed above

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +95,35 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
-        #     # Create BoundingBox object and append to list
-        #     result, box = ...
-
-        return [], image_annotated
+        for raw_box in boxes_cpu:
+            # Create BoundingBox object and append to list
+            result, box = bounding_box.BoundingBox.create(raw_box)
+            # skip if BoundingBox creatation failed
+            if result is None:
+                pass
```
TongguangZhang, line 123: I don't think pass is what you wanna do here - maybe continue?

however, instead of skipping failed boxes, we acc wanna return an empty list - in data science, when part of a data point is corrupted, we skip the whole data point instead of returning partial data, in this case, 1 image is 1 data point, and the boxes are part of the image, so we skip the whole image if any of the boxes fail

TongguangZhang (commented):
reviewed

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,10 +66,27 @@ def run(
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
-        # Do something based on the report and the state of this class...
+        # unpack report attributes
+        status = report.status
+        position = report.position
```
TongguangZhang, line 71: i usually recommend not doing this - it's more clear when u use the full variable since we know that it's the report's status, position, etc.

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,10 +66,27 @@ def run(
...
+        if status == drone_status.DroneStatus.HALTED and (
+            self.acceptance_radius**2 >= find_squared_dist(self.waypoint, position)
+        ):
+            command = command.create_land_command()
+        elif status == drone_status.DroneStatus.HALTED:
```
TongguangZhang, line 76: you check if the drone is halted twice here, maybe do a nested if?

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,10 +66,27 @@ def run(
...
 
         # ============
         # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
         # ============
 
         return command
+
+
+# helper functions
+def find_squared_dist(loc1: location.Location, loc2: location.Location) -> int:
```
TongguangZhang, line 90: you should declare this inside the class (I also recommend making it a static method)

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +69,57 @@ def run(
...
         return command
+
+
+# helper functions
+def find_squared_dist(loc1: location.Location, loc2: location.Location) -> int:
+    """helper function that finds the squared distance between two Location instances"""
+    return (loc1.location_x - loc2.location_x) ** 2 + (loc1.location_y - loc2.location_y) ** 2
+
+
+def find_closest_location(
```
TongguangZhang, line 113: declare these inside the class

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +69,57 @@ def run(
...
+        ):
+            # landing at closest landing pad
+            command = command.create_land_command()
+        elif status == drone_status.DroneStatus.HALTED and (
+            self.acceptance_radius**2 >= find_squared_dist(self.waypoint, position)
+        ):
+            # find closest landing pad
+            self.waypoint_achieved = True
+            final_destination = find_closest_location(position, landing_pad_locations)
+            self.final_destination = final_destination
```
TongguangZhang, line 89: maybe make this one line, you don't really need the second statement

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +69,57 @@ def run(
...
+    """find the closest location in a list of locations"""
+    min_dist, closest_location = float("inf"), None
+
+    for loc in location_list:
+        dist = find_squared_dist(curr_loc, loc)
+        if dist < min_dist:
+            closest_location = loc
+            min_dist = dist
+
+    return closest_location
```
TongguangZhang, line 125: what if the list is empty and closest location is None? make sure to handle that in the run function

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +95,36 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
         # * conf
         # * device
         # * verbose
-        predictions = ...
+        predictions = self.__model.predict(source=image, verbose=False, conf=0.25, device=None)
```
TongguangZhang, line 98: 0.25 might be a little low for conf

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +95,36 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         bounding_boxes = []
         # Hint: .shape gets the dimensions of the numpy array
-        # for i in range(0, ...):
-        #     # Create BoundingBox object and append to list
-        #     result, box = ...
+        for raw_box in boxes_cpu:
+            # Create BoundingBox object and append to list
+            result, box = bounding_box.BoundingBox.create(raw_box)
+            # discard image if failed
+            if result is None:
```
TongguangZhang, line 122: result won't be none, it's a boolean (make sure to check the create function to understand what it returns)

TongguangZhang (commented):
reviewed

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +95,36 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
         # * conf
         # * device
         # * verbose
-        predictions = ...
+        predictions = self.__model.predict(source=image, verbose=False, conf=0.75, device=None)
```
maxlou05, line 98: Although using default or letting it choose is fine, there is a variable in this bootcamp that indicates which device you should use (You don't have to change anything here, just something to think about)

maxlou05 (commented):
Conditionally approved!

PR #142, opened 2025-01-13
Teighan Miller - Bootcamp

teighanmiller:


modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,6 +66,25 @@ def run(
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
+        # If the drone is halted and not at the destination, move the drone to destination
+        if report.status == drone_status.DroneStatus.HALTED and report.destination != self.waypoint:
```
TongguangZhang, line 70: use the acceptance radius, not != (float operations are a little funky, so you can't do a direct comparison like this)

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,6 +66,25 @@ def run(
...
+            command = commands.Command.create_set_relative_destination_command(
+                self.waypoint.location_x, self.waypoint.location_y
+            )
```
TongguangZhang, line 73: note that it's relative distance, not absolute, so you can't just use the coords of the waypoint

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,6 +66,25 @@ def run(
...
+        # if the drone is at the destination and halted land the drone
+        elif (
+            report.status == drone_status.DroneStatus.HALTED and report.destination == self.waypoint
+        ):
```
TongguangZhang, line 77: same as above - use acceptance radius instead of ==

TongguangZhang, line 77: it seems like you make a redundant calculation here - you can nest the if statements, which lets u do if else
```
if halted:
    if at location:
        ...
    else:
        ...

File: modules/bootcamp/decision_simple_waypoint.py [hunk 4]
```diff
@@ -68,6 +66,25 @@ def run(
...
+            command = commands.Command.create_land_command()
+
+        # Case handling to ensure that the drone tries to land again if it lands outside of the acceptance radius
+        if (
+            report.status == drone_status.DroneStatus.LANDED
+            and (report.position - report.destination) > self.acceptance_radius
+        ):
+            command = commands.Command.create_set_relative_destination_command(
+                self.waypoint.location_x, self.waypoint.location_y
+            )
```
TongguangZhang, line 87: it doesn't make sense to land the drone and then take off again if it's in the wrong spot - just make sure it's in the right spot before landing

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -4,6 +4,7 @@
 Travel to designated waypoint and then land at a nearby landing pad.
 """
 
+from math import sqrt
```
TongguangZhang, line 7: note that the bootcamp says to not import math - think about how you would do it without the math lib

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +70,59 @@ def run(
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
+        if not self.target and report.status == drone_status.DroneStatus.HALTED:
```
TongguangZhang, line 73: usually better to check with `self.target is None`

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +70,59 @@ def run(
...
+            self.target = self.distance(report.position, landing_pad_locations)
+
+        # If the drone is halted and not at the destination, move the drone to destination
+        if report.status == drone_status.DroneStatus.HALTED and report.destination != self.target:
```
TongguangZhang, line 77: acceptance radius, not !=

modules/bootcamp/decision_waypoint_landing_pads.py
TongguangZhang, line 73: acceptance radius, not ==

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +70,59 @@ def run(
...
+
+            command = commands.Command.create_set_relative_destination_command(
+                self.target.location_x, self.target.location_y
+            )
+        # if the drone is at the destination and halted land the drone
+        elif report.status == drone_status.DroneStatus.HALTED and report.destination == self.target:
```
TongguangZhang, line 83: acceptance radius, not ==

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +70,59 @@ def run(
...
+
+        # Case handling to ensure that the drone tries to land again if it lands outside of the acceptance radius
+        if (
+            report.status == drone_status.DroneStatus.LANDED
+            and (report.position - report.destination) > self.acceptance_radius
+        ):
+            command = commands.Command.create_set_relative_destination_command(
+                self.target.location_x, self.target.location_y
+            )
+
```
TongguangZhang, line 94: same as above, don't land until the drone is in the right spot instead of taking off again

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +70,59 @@ def run(
...
         # Do something based on the report and the state of this class...
 
         # ============
         # ↑ BOOTCAMPERS MODIFY ABOVE THIS COMMENT ↑
         # ============
 
         return command
+
+    def distance(
```
TongguangZhang, line 103: maybe name this something more descriptive

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +70,59 @@ def run(
...
+        closest destination. This location.Location object is returned.
+
+        Args:
+            current_location (location.Location): Location the drone is currently at (0, 0)
+            destination_list (list[location.Location]): list of possible destinations for the drone
+
+        Returns:
+            location.Location: class implementation of a location defined by x and y
+        """
+        destination = location.Location(1000000000, 1000000000)
```
TongguangZhang, line 117: use infinity instead of a large number

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +70,59 @@ def run(
...
+        for landing_pad in destination_list:
+            if sqrt(
+                (landing_pad.location_x - current_location.location_x) ** 2
+                + (landing_pad.location_y - current_location.location_y) ** 2
+            ) < sqrt(
+                (destination.location_x - current_location.location_x) ** 2
+                + (destination.location_y - current_location.location_y) ** 2
+            ):
```
TongguangZhang, line 125: do you need sqrt?

hint: does a < b => a^2 < b^2 in this case?

modules/bootcamp/decision_waypoint_landing_pads.py
TongguangZhang, line 103: does this function need self? consider making it a static method

modules/bootcamp/decision_waypoint_landing_pads.py
TongguangZhang, line 125: note that if there are no landing pads in the list, you'll send a command that makes the drone fly to (1000000000, 1000000000), which is not what we want

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +98,38 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
 
         # Get the xyxy boxes list from the Boxes object in the Result object
-        boxes_xyxy = ...
+        boxes_xyxy = prediction.boxes.xyxy
 
         # Detach the xyxy boxes to make a copy,
         # move the copy into CPU space,
         # and convert to a numpy array
-        boxes_cpu = ...
+        boxes_cpu = prediction.boxes.xyxy.cpu().numpy()
```
TongguangZhang, line 116: make sure to call detach(), this removes it from the computational graph, which makes it stop tracking the gradients
basically, during training, gradients are tracked on the tensors for back propagation, but since we're only using inference, we don't need the gradients, and they take up a lot of memory, so we should stop tracking them

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +98,38 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
 
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
+        for i in range(0, boxes_cpu.shape[0]):
+            new_box = bounding_box.BoundingBox.create(boxes_xyxy[i])
```
TongguangZhang, line 121: better practice to do `result, new_box = ...`

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +98,38 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+            if new_box[0]:
+                bounding_boxes += new_box
```
TongguangZhang, line 123: instead of skipping failed boxes, we acc wanna return an empty list - in data science, when part of a data point is corrupted, we skip the whole data point instead of returning partial data, in this case, 1 image is 1 data point, and the boxes are part of the image, so we skip the whole image if any of the boxes fail

TongguangZhang, line 123: use .append for lists - also, do you need to add the success flag to the bounding box list instead of just the box

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +98,38 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Hint: .shape gets the dimensions of the numpy array
         # for i in range(0, ...):
         #     # Create BoundingBox object and append to list
         #     result, box = ...
 
-        return [], image_annotated
+        if not bounding_boxes:
+            return [], image_annotated
```
TongguangZhang, line 130: this can be deleted if you don't add the success flags

TongguangZhang (commented):
reviewed

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,6 +66,16 @@ def run(
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
+        # If the drone is halted and not at the destination, move the drone to destination
+        if report.status == drone_status.DroneStatus.HALTED:
+            if (report.destination - self.waypoint) < self.acceptance_radius:
```
maxlou05, line 71: Please use Euclidean distance and not this. I'm not even sure how this comparison works, as you are comparing a vector (in this case a class with 2 ints) with a scalar (1 int).

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -38,6 +38,8 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
         # ============
 
         # Add your own
+        self.target = None
+        self.tol = 0.0001
```
maxlou05, line 42: The tolerance is given to you as `acceptance_radius`

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,6 +66,16 @@ def run(
...
+                command = commands.Command.create_set_relative_destination_command(
+                    abs(self.waypoint.location_x), abs(self.waypoint.location_y)
```
maxlou05, line 73: It is a relative distance command, so your own position matters. Also, don't use abs because (-1, -1) is a valid location and is different from (1, 1)

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +70,62 @@ def run(
...
         # ============
 
+        if not self.target and report.status == drone_status.DroneStatus.HALTED:
+            self.target = self.find_closest_landing_pad(report.position, landing_pad_locations)
+
+        # If the drone is halted and not at the destination, move the drone to destination
+        if report.status == drone_status.DroneStatus.HALTED and report.destination != self.target:
+
+            command = commands.Command.create_set_relative_destination_command(
+                self.target.location_x, self.target.location_y
```
maxlou05, line 80: Relative distance command, your position matters

modules/bootcamp/decision_waypoint_landing_pads.py
maxlou05: Instead of doing this extra case, can you just add to the condition of the elif above? This doesn't really work right now because the drone doesn't accept move commands after it lands, so incorporate it into the one above

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +70,62 @@ def run(
...
+    @staticmethod
+    def find_closest_landing_pad(
+        current_location: location.Location, destination_list: "list[location.Location]"
+    ) -> location.Location:
+        """
+        Takes the current location and a list of possible destinations and calculates the
+        closest destination. This location.Location object is returned.
+
+        Args:
+            current_location (location.Location): Location the drone is currently at (0, 0)
```
maxlou05, line 112: It starts at (0,0), but you cannot assume the drone doesn't move (you don't know where the drone is when you call this function). However, it doesn't really matter so it's fine. Returning to (0,0) is ok if there are no landing pads, great job for thinking about this case! (You do not need to change anything here)

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +98,40 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Detach the xyxy boxes to make a copy,
         # move the copy into CPU space,
         # and convert to a numpy array
-        boxes_cpu = ...
+        boxes_cpu = prediction.boxes.xyxy.cpu().detach().numpy()
 
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
+        for i in range(0, boxes_cpu.shape[0]):
+            result, new_box = bounding_box.BoundingBox.create(boxes_xyxy[i])
```
maxlou05, line 121: Please use `boxes_cpu` here

maxlou05 (commented):
Reviewed

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,10 +69,62 @@ def run(
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
+        if self.target is None and report.status is drone_status.DroneStatus.HALTED:
+            self.target = self.find_closest_landing_pad(report.position, landing_pad_locations)
```
maxlou05, line 73: For task 4, you need to fly to the waypoint first (you can think of it as target1), before you can fly to the nearest landing form the waypoint (target2). So in total you need to go to 2 things, but you only go to 1 (the landing pad)

maxlou05 (commented):
Reviewed. Also check discord about the distance thing

maxlou05 (commented):
Conditionally approved! I would advise to be careful when using is, because I think that is just compares pointers and not compares objects (does not invoke the object's __eq__()). I think it works in your case since most object assignments are just references. This is just an FYI for future.

PR #146, opened 2025-01-21
Sophia's Bootcamp

soappeea:


modules/bootcamp/.DS_Store
AaronWang04, line 1: careful about doing git add . in the future

I would learn the other options such as git add -u which only stages modified files, not new files.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -38,11 +38,22 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
...
         # ============
 
+    def dist_between(self, final_location: location.Location) -> float:
+        """
+        Calculate the squared distance between the waypoint and a given location.
+        """
+        return (self.waypoint.location_x - final_location.location_x) ** 2 + (
+            self.waypoint.location_y - final_location.location_y
+        ) ** 2
+
```
AaronWang04, line 56: good use of typing

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -37,8 +37,10 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
-        # Add your own
+        self.min_flight_bound = -60
+        self.max_flight_bound = 60
```
AaronWang04, line 41: I don't think these are necessary but it doesn't hurt to have an extra check

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +95,37 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
         # Hint: .shape gets the dimensions of the numpy array
         # for i in range(0, ...):
         #     # Create BoundingBox object and append to list
         #     result, box = ...
+        for i in range(boxes_cpu.shape[0]):
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+            if result:
+                bounding_boxes.append(box)
```
AaronWang04, line 126: an alternate way (more common practice) of dealing with corrupt output is to just disregard the whole image and return nothing

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,6 +80,42 @@ def run(
...
+            and self.min_flight_bound <= self.waypoint.location_y <= self.max_flight_bound
+        ):
+            distance_from_waypoint = (
+                self.waypoint.location_x - report.position.location_x
+            ) ** 2 + ((self.waypoint.location_y - report.position.location_y) ** 2)
+
+            if report.status == drone_status.DroneStatus.HALTED:
+
+                if self.ready_to_land:
+                    command = commands.Command.create_land_command()
```
AaronWang04, line 94: the acceptance radius is to counter the analog error of a GPS in real life, you should check it both on 1. reaching initial waypoint and 2. landing pad location

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,6 +80,42 @@ def run(
...
+                    smallest_distance = float("inf")
+                    closest_pad = None
+
+                    for landing_pad in landing_pad_locations:
+                        if self.dist_between(landing_pad) < smallest_distance:
+                            smallest_distance = self.dist_between(landing_pad)
+                            closest_pad = landing_pad
+
+                    if closest_pad is None:
+                        command = commands.Command.create_null_command()
```
AaronWang04, line 106: I wouldn't do null into landing if no landing pad was found. A better implementation is to let the drone stay halted until the function was called with proper landing_pad_locations

AaronWang04 (commented):
some minor issues, overall well done!

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -69,6 +82,46 @@ def run(
...
+            ) ** 2 + ((self.waypoint.location_y - report.position.location_y) ** 2)
+
+            if report.status == drone_status.DroneStatus.HALTED:
+
+                if (
+                    self.waypoint_reached
+                    and self.closest_pad is not None
+                    and (self.closest_pad.location_x - report.position.location_x) ** 2
+                    + (self.closest_pad.location_y - report.position.location_y) ** 2
+                    < self.acceptance_radius**2
```
AaronWang04, line 100: you can probably make the above dist_between function more general (between two points) to make this less messy

AaronWang04 (approved):
left one comment but LGTM overall

PR #147, opened 2025-01-22
Bootcamp - Han Zhou

Han-Zhou:


modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -4,6 +4,7 @@
 Travel to designated waypoint.
 """
 
+from math import sqrt
```
maxlou05, line 7: The bootcamp is doable without importing anything. Think about how you can not use the square root function (Hint: think about the purpose for which you are calculating distances)

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -68,7 +69,32 @@ def run(
...
+        distance_from_waypoint = sqrt(
+            (self.waypoint.location_x - position.location_x) ** 2
+            + (self.waypoint.location_y - position.location_y) ** 2
+        )
+
+        # if the drone is in the acceptance radius, we have reached
+        if distance_from_waypoint < self.acceptance_radius:
+            self.reached = True
+
+        if self.reached:
```
maxlou05, line 84: This works, but you can probably just get rid of the first condition because they are the same. Also, I would suggest checking drone status first before you check distance, which saves a little bit of compute considering the drone is moving most of the time

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +96,33 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         bounding_boxes = []
         # Hint: .shape gets the dimensions of the numpy array
-        # for i in range(0, ...):
-        #     # Create BoundingBox object and append to list
-        #     result, box = ...
+        for i in range(0, boxes_cpu.shape[0]):
+            # Create BoundingBox object and append to list
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+            if result:
+                bounding_boxes.append(box)
```
maxlou05, line 123: In statistics, we generally discard the entire data point (the whole image) if one part of it (a bounding box) goes wrong. (You do not need to change anything here, just something to think about)

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -4,6 +4,8 @@
 Travel to designated waypoint and then land at a nearby landing pad.
 """
 
+from math import sqrt
```
maxlou05, line 7: Same thing, doable without square root

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -68,7 +91,58 @@ def run(
...
+            distance_from_waypoint = sqrt(
+                (self.waypoint.location_x - position.location_x) ** 2
+                + (self.waypoint.location_y - position.location_y) ** 2
+            )
+
+            # if the drone is in the acceptance radius, we have reached
+            if distance_from_waypoint < self.acceptance_radius:
+                self.reached = True
+
+            if self.reached:
```
maxlou05, line 107: Same, you can combine these

maxlou05 (commented):
Reviewed

maxlou05 (commented):
Approved!

PR #151, opened 2025-02-01
Krish's Bootcamp

krish1905:


modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +98,38 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
         # Hint: .shape gets the dimensions of the numpy array
         # for i in range(0, ...):
         #     # Create BoundingBox object and append to list
         #     result, box = ...
-
-        return [], image_annotated
+        for i in range(boxes_cpu.shape[0]):
+            x1, y1, x2, y2 = boxes_cpu[i, 0:4]
```
maxlou05, line 127: This should be fine, but fyi boxes_cpu comes from boxes_xyxy, which have just the xyxy information. You don't have to re-extract the information and then put it back in a list. (You don't have to change anything here)

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +98,38 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+            bounds = np.array([x1, y1, x2, y2], dtype=float)
+            result, box = bounding_box.BoundingBox.create(bounds)
+            if result:
+                bounding_boxes.append(box)
```
maxlou05, line 131: Normally in statistics, we discard the entire data point (the whole image) if a part of it (a bounding box) goes wrong. (You don't have to change anything here, just something to think about)

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -61,14 +62,60 @@ def run(
             put_output(command)
         ```
         """
-        # Default command
+        # Default comment
```
maxlou05, line 65: Not sure the purpose of this change? It is the default command

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -61,14 +62,60 @@ def run(
...
+        if self._phase == "WAYPOINT":
+            if (
+                dist_sq(report.position, self.waypoint)
+                <= self.acceptance_radius * self.acceptance_radius
+            ):
+                if report.status == drone_status.DroneStatus.HALTED:
+                    closest_dist_sq = float("inf")
+                    chosen_pad = None
+                    for pad in landing_pad_locations:
+                        pad_dist = dist_sq(self.waypoint, pad)
```
maxlou05, line 89: maybe also add the _sq to this variable name, as it looks like you're comparing a non-squared distance with a squared distance

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -61,14 +62,60 @@ def run(
...
+            dx = loc1.x - loc2.x
+            dy = loc1.y - loc2.y
+            return dx * dx + dy * dy
+
+        if self._phase == "WAYPOINT":
+            if (
+                dist_sq(report.position, self.waypoint)
+                <= self.acceptance_radius * self.acceptance_radius
+            ):
+                if report.status == drone_status.DroneStatus.HALTED:
```
maxlou05, line 85: You can add optionally add an else to this if and make the drone halt if it has arrived at the waypoint/desired location just to make it clearer, or add a comment saying why you never halt the drone. (You don't have to change anything here)

maxlou05 (commented):
Reviewed

maxlou05 (commented):
Approved!

PR #153, opened 2025-03-05
completed tasks 1,2,3 and 4

Diyako1:


modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -69,7 +71,38 @@ def run(
...
+        if not self._has_set_destination:
+            if status == drone_status.DroneStatus.HALTED:
+                new_x = dist_x
+                new_y = dist_y
+
+                final_x = current_pos.x + new_x
+                final_y = current_pos.y + new_y
+                if not (-60.0 <= final_x <= 60.0 and -60.0 <= final_y <= 60.0):
+                    pass
+                command = commands.Command.create_set_relative_destination_command(new_x, new_y)
```
maxlou05, line 102: This is a relative position command, so you don't have to add back your position. It doesn't go to an absolute location

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -69,7 +71,38 @@ def run(
...
+                self._has_set_destination = True
+        else:
+            pass
```
maxlou05, line 105: What if the drone set the destination, but then failed to get there due to some circumstances? How can you guarantee that it gets there?

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -61,17 +55,72 @@ def run(
...
+                        current_pos, landing_pad_locations
+                    )
+                else:
+                    self._landing_pad_location = self.waypoint
+            distance_to_pad = self._distance(current_pos, self._landing_pad_location)
+            if distance_to_pad <= self.acceptance_radius:
+                self._phase = 2
+                self._has_set_destination = False
+            else:
+                command = self._move_toward(report, self._landing_pad_location)
```
maxlou05, line 85: You put has_set_destination in the wrong branch?

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -61,17 +55,72 @@ def run(
...
+        if status == drone_status.DroneStatus.LANDED:
+            return command
+
+        if self._phase == 0:
+            distance_to_waypoint = self._distance(current_pos, self.waypoint)
+            if distance_to_waypoint <= self.acceptance_radius:
+                self._phase = 1
+                self._has_set_destination = False
+            else:
+                command = self._move_toward(report, self.waypoint)
```
maxlou05, line 71: has_set_destination is in the wrong branch?

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -61,17 +55,72 @@ def run(
...
+                    self._has_landed = True
+                elif status != drone_status.DroneStatus.HALTED:
+                    command = commands.Command.create_halt_command()
 
         return command
+
+    def _distance(self, loc1: location.Location, loc2: location.Location) -> float:
+        dy = loc2.y - loc1.y
+        dx = loc2.x - loc1.x
+        return (dx**2 + dy**2) ** 0.5
```
maxlou05, line 102: Try to avoid using the square root function

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +98,38 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         bounding_boxes = []
         # Hint: .shape gets the dimensions of the numpy array
         # for i in range(0, ...):
         #     # Create BoundingBox object and append to list
         #     result, box = ...
-
-        return [], image_annotated
+        for i in range(boxes_cpu.shape[0]):
+            x1, y1, x2, y2 = boxes_cpu[i, 0:4]
+            bounds = np.array([x1, y1, x2, y2], dtype=float)
```
maxlou05, line 128: I don't think you need to unpack and then re-pack the array, you should be able to just directly put in boxes_cpu[i] into BoundingBox.create

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +98,38 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+            result, box = bounding_box.BoundingBox.create(bounds)
+            if result:
+                bounding_boxes.append(box)
```
maxlou05, line 131: Normally in statistics, we normally discard the entire data point (the whole image) if part of it (a bounding box) goes wrong. You do not need to change anything here, just something to think about.

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -61,17 +55,74 @@ def run(
...
+    def _move_toward(
+        self, report: drone_report.DroneReport, target: location.Location
+    ) -> commands.Command:
+        if self._has_set_destination:
+            return commands.Command.create_null_command()
+        if report.status == drone_status.DroneStatus.HALTED:
+            dx = target.x - report.position.x
+            dy = target.y - report.position.y
+            self._has_set_destination = True
+            return commands.Command.create_set_relative_dest_command(dx, dy)
```
RohanKatreddy, line 127: this is not a valid command, replace dest with destination

RohanKatreddy (changes requested):
Reviewed

RohanKatreddy (approved):
lgtm!

PR #159, opened 2025-04-13
Completed BootCamp - Sophie Yuen

happyface3650:


modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,22 +98,23 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
         # * conf
         # * device
         # * verbose
-        predictions = ...
+
+        predictions = self.__model.predict(image, conf=0.5, device=self.__DEVICE, verbose=False)
```
ellyokes253, line 102: nitpick, but you'll wanna set `source=image` instead of just having image there

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,22 +98,23 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
 
         # Get the Result object
-        prediction = ...
+        prediction = predictions[0]
 
         # Plot the annotated image from the Result object
         # Include the confidence value
-        image_annotated = ...
+        image_annotated = prediction.plot()
```
ellyokes253, line 109: include the confidence value in the parameters

modules/bootcamp/detect_landing_pad.py
```diff
@@ -122,7 +123,14 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
         #     # Create BoundingBox object and append to list
         #     result, box = ...
 
-        return [], image_annotated
+        for bounds in boxes_cpu:
+            bounds = np.asarray(bounds).reshape(4)
```
ellyokes253, line 127: although this is correct, there's no need to reconvert this into an array because when you declare `boxes_cpu`, you already make it into a numpy array. therefore there's also no need to reshape either

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,22 +98,25 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
+        )
 
         # Get the Result object
-        prediction = ...
+        prediction = predictions[0]
 
         # Plot the annotated image from the Result object
         # Include the confidence value
-        image_annotated = ...
+        image_annotated = prediction.plot(conf=0.5)
```
ellyokes253, line 111: for the confidence parameter is a bool and not a number: https://docs.ultralytics.com/modes/predict/#plot-method-parameters

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -67,9 +65,24 @@ def run(
...
         # Do something based on the report and the state of this class...
-
+        if (current_position.location_x == 0) and (current_position.location_y == 0):
+            command = commands.Command.create_set_relative_destination_command(
+                self.waypoint.location_x, self.waypoint.location_y
+            )
+        elif (
+            (current_position.location_x == waypoint.location_x)
+            and (current_position.location_y == waypoint.location_y)
+            and (report.status != drone_status.DroneStatus.HALTED)
```
ellyokes253, line 78: this is where you're gonna want to use that `self.acceptance_radius` to see if the drone is within that to be able to send a landing command

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -67,9 +65,24 @@ def run(
         # ============
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
-
+        current_position = report.position
+        waypoint = self.waypoint
```
ellyokes253, line 69: nitpick: if you're not using this variable no need to set it

modules/bootcamp/decision_simple_waypoint.py
```diff
@@ -67,9 +65,24 @@ def run(
...
         # Do something based on the report and the state of this class...
-
+        if (current_position.location_x == 0) and (current_position.location_y == 0):
```
ellyokes253, line 71: while this may work as initially the drone's relative position when starting is (0,0), think about the fact that this `run()` function is called several times in the main code and this will only be true at the very beginning

modules/bootcamp/decision_simple_waypoint.py
happyface3650: but i am using it in the if statements as arguments to the _reached function?

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -67,8 +94,53 @@ def run(
         # ============
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
-
-        # Do something based on the report and the state of this class...
+        current_position = report.position
+        min_distance = 28800
```
ellyokes253, line 98: instead of using a finite large value, try using `float("inf")`

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -67,8 +94,53 @@ def run(
...
+            distance = self._distance(self.waypoint, pad)
+            if distance[0] < min_distance:
+                min_distance = distance[0]
+                x_distance = distance[1]
+                y_distance = distance[2]
+                closest_pad = pad
+
+        if (
+            (current_position.location_x == 0)
+            and (current_position.location_y == 0)
```
ellyokes253, line 112: again, similar to the thought of this only being true once. i think using the halted drone status along with the step flag should be sufficient for what you're trying to achieve

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -67,8 +94,53 @@ def run(
...
+        elif (self.step == "WAYPOINT") and self._reached(current_position, self.waypoint):
+            command = commands.Command.create_halt_command()
+            self.step = "STOPPED_AT_WAYPOINT"
+        elif (
+            (self.step == "STOPPED_AT_WAYPOINT")
+            and self._reached(current_position, self.waypoint)
+            and (report.status == drone_status.DroneStatus.HALTED)
+        ):
+            command = commands.Command.create_set_relative_destination_command(
+                x_distance, y_distance
```
ellyokes253, line 129: are you sure that the closest landing pad that was calculated before the drone started moving is the same as the one closest to the waypoint?

modules/bootcamp/decision_simple_waypoint.py
ellyokes253: you use `self.waypoint` instead of just using waypoint like you declared

modules/bootcamp/decision_waypoint_landing_pads.py
happyface3650: Yes, because I used waypoint coordinates while calculating the distance and determining the closest landing pad, rather than the starting position. Please let me know if this is wrong.

modules/bootcamp/decision_simple_waypoint.py
happyface3650: ok I changed all of them to waypoint for better clarity

modules/bootcamp/decision_waypoint_landing_pads.py
ellyokes253: oh lmao i missed that ur good

ellyokes253 (commented):
lgtm!!

PR #161, opened 2025-04-24
Sirtaj Singh - Autonomy Bootcamp

SirSingh:


ellyokes253:
reviewed

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +98,35 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         bounding_boxes = []
-        # Hint: .shape gets the dimensions of the numpy array
-        # for i in range(0, ...):
-        #     # Create BoundingBox object and append to list
-        #     result, box = ...
+        # Iterate over all rows of boxes
+        for row in boxes_cpu:
+            # Create  box object for each xyxy
+            box = bounding_box.BoundingBox.create(row)[1]
+            bounding_boxes.append(box)
```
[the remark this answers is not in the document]
SirSingh: Updated to only add box to list if it is not None

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +98,36 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
-        # for i in range(0, ...):
-        #     # Create BoundingBox object and append to list
-        #     result, box = ...
 
-        return [], image_annotated
+        bounding_boxes = []
+        # Iterate over all rows of boxes
+        for row in boxes_cpu:
+            # Create  box object for each xyxy
+            box = bounding_box.BoundingBox.create(row)[1]
```
ellyokes253, line 126: `bounding_box.BoundingBox.create()` returns a tuple of a boolean and the box, and you should use the boolean in your logic

SirSingh: Updated to use boolean value

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +98,36 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
-        #     result, box = ...
 
-        return [], image_annotated
+        bounding_boxes = []
+        # Iterate over all rows of boxes
+        for row in boxes_cpu:
+            # Create  box object for each xyxy
+            valid_box, box = bounding_box.BoundingBox.create(row)
+            if valid_box:
+                bounding_boxes.append(box)
```
ellyokes253, line 128: in stats, if there was ever an invalid data point in the set, we would discard the whole set instead of leave out the invalid point, in other words, return `[], image_annotated` (you don't need to fix anything here, just something to think about)

ellyokes253:
lgtm!

PR #162, opened 2025-05-03
Daniel Zhao Autonomy Bootcamp

danielquzhao:


modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +95,37 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         # Loop over the boxes list and create a list of bounding boxes
         bounding_boxes = []
         # Hint: .shape gets the dimensions of the numpy array
         # for i in range(0, ...):
         #     # Create BoundingBox object and append to list
         #     result, box = ...
+        for i in range(0, boxes_cpu.shape[0]):
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+            if result:
+                bounding_boxes.append(box)
```
RohanKatreddy, line 126: So instead of skipping failed boxes, we acc wanna return an empty list - in data science, when part of a data point is corrupted, we skip the whole data point instead of returning partial data, in this case, 1 image is 1 data point, and the boxes are part of the image, so we skip the whole image if any of the boxes fail

modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +95,43 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
...
         #     # Create BoundingBox object and append to list
         #     result, box = ...
-
-        return [], image_annotated
+        boxes_valid = True
+        for i in range(0, boxes_cpu.shape[0]):
+            result, box = bounding_box.BoundingBox.create(boxes_cpu[i])
+            if not result:
+                boxes_valid = False
+                break
```
RohanKatreddy, line 128: So here you can just return rather than updating a variable and postponing the return statement

RohanKatreddy (approved):
Great job!

PR #168, opened 2025-06-11
Ananth Malur Anand's bootcamp

ananth-2022:


modules/bootcamp/detect_landing_pad.py
```diff
@@ -98,31 +95,31 @@ def run(self, image: np.ndarray) -> "tuple[list[bounding_box.BoundingBox], np.nd
         # * conf
         # * device
         # * verbose
-        predictions = ...
-
+        predictions = self.__model.predict(image, conf=0.7)
```
ellyokes253, line 98: you're missing a few parameters that predictions needs, not just the confidence value

modules/bootcamp/decision_waypoint_landing_pads.py
```diff
@@ -37,7 +34,8 @@ def __init__(self, waypoint: location.Location, acceptance_radius: float) -> Non
         # ↓ BOOTCAMPERS MODIFY BELOW THIS COMMENT ↓
         # ============
 
-        # Add your own
+        self.reached_waypoint = False
+        self.target_landing_pad = location.Location(9999, 9999)
```
ellyokes253, line 38: instead of setting this to an arbitrary location value, you could set this as something more relevant

ellyokes253:
reviewed!

PR #173, opened 2025-07-03
Feature/task 3 and 4 edited

2sunz:
Here I fixed the distance calculation by comparing squared distances instead of actual distances. I also deleted the extra  retunr command lines as instructed.


ellyokes253:
lgtm!

